PR #2, opened 2021-04-08
Docs and gosec fixes, rename test to preview

marcinwyszynski:


README.md
```diff
@@ -1,2 +1,188 @@
...
+CATEGORY:
+   Run management
+
+OPTIONS:
+   --sha value  Commit SHA for the newly created run
+   --tail       Indicate whether to tail the run (default: false)
+   --help, -h   show help (default: false)
+```
+
+This subcommand creates a [tracked run](https://docs.spacelift.io/concepts/run/tracked) on a stack. By default, it uses the latest tracked commit, but you can also point it to an arbitrary SHA using the `--sha` flag. Also by default, this command only creates a run and prints out the URL where the run can be accessed. If the `--tail` flag is set however, run logs will be retrieved until the run terminates. If the run is tailed, the exit code of the command will depend on the outcome of the run - if the run is `FINISHED`, the command succeeds (exits with `0`), otherwise it fails.
```
adamconnelly, line 114: This is maybe a wider question than just for the CLI, but would it make sense to stick to one name (i.e. either _deployment_ or _tracked run_)?

While both names are used, it definitely makes sense to explain it here like you've done, so I'm not suggesting you change this text just now. I just bring it up because it feels slightly weird to `spacelift-cli stack deploy` if we then say it creates a "tracked run" rather than saying it "deploys the stack".

README.md
```diff
@@ -1,2 +1,188 @@
...
+
+Notes:
+
+- terminating the executable (Ctrl+C) while tailing logs does not stop the run, it just stops tailing logs;
```
adamconnelly, line 118:
```suggestion
- Terminating the executable (Ctrl+C) while tailing logs does not stop the run, it just stops tailing logs.
```

README.md
```diff
@@ -1,2 +1,188 @@
...
+
+- you cannot confirm or discard the run from the command line. If the deployment ends up in [`UNCONFIRMED`](https://docs.spacelift.io/concepts/run/tracked#unconfirmed) state, the CLI be stuck until the state machine progresses. You will need to go to the run URL and make a decision there;
```
adamconnelly, line 120:
```suggestion
- You cannot confirm or discard the run from the command line. If the deployment ends up in [`UNCONFIRMED`](https://docs.spacelift.io/concepts/run/tracked#unconfirmed) state, the CLI be stuck until the state machine progresses. You will need to go to the run URL and make a decision there.
```

README.md
```diff
@@ -1,2 +1,188 @@
...
+
+CATEGORY:
+   Run management
+
+OPTIONS:
+   --run value  ID of the run
+   --help, -h   show help (default: false)
+```
+
+This subcommand prints out (and tails) logs for an existing run or task. The exit code of the command will depend on the outcome of the run - if the run is `FINISHED`, the command succeeds (exits with `0`), otherwise it fails.
```
adamconnelly, line 161: This feels a little weird in the sense that the `logs` command may have succeeded, but we'll return a non-zero exit code depending on the state of the run. I wonder if a separate command to get the run state would make more sense?

Did you have any use-case in mind for this, or was it just arbitrary?

README.md
```diff
@@ -1,2 +1,188 @@
...
+CATEGORY:
+   Run management
+
+OPTIONS:
+   --noinit    Indicate whether to skip initialization for a task (default: false)
+   --tail      Indicate whether to tail the run (default: false)
+   --help, -h  show help (default: false)
+```
+
+This subcommand starts a [task](https://docs.spacelift.io/concepts/run/task) against a stack. The command for the task itself goes after the CLI bit. It can be quoted to prevent any confusion in from shell tokenization. Example:
```
adamconnelly, line 184:
```suggestion
This subcommand starts a [task](https://docs.spacelift.io/concepts/run/task) against a stack. The command for the task itself must be specified after all the other `spacelift-cli` command arguments. It can be quoted to prevent any confusion from shell tokenization. Example:
```

README.md
```diff
@@ -1,2 +1,188 @@
...
+```bash
+curl -s -L https://github.com/spacelift-io/spacelift-cli/releases/latest/download/spacelift-cli-darwin-amd64.tar.gz | tar xz -C /tmp && \
+mv /tmp/spacelift-cli-darwin-amd64 /usr/local/bin/spacelift-cli
+```
+
+## Authentication
+
+Spacelift CLI is designed to work in two different contexts - a non-interactive scripting mode (eg. external CI/CD pipeline) and local interactive mode, meaning you typing things in your shell. It thus supports two types of credentials - environment variables and user profiles. Let's call these "credential providers", like AWS does. We'll look into them separately.
+
+### Authenticating using environment variables
```
adamconnelly, line 18: @marcinwyszynski this is very subjective, so feel free to ignore me here. What you've written is fine - I just felt like it could be easier to quickly understand the options if we split things up a bit into a section for each authentication method, and were explicit in each section about exactly what variables are required.

Here's my suggested rewording:

```markdown
### Authenticating using environment variables

The spacelift CLI supports the following authentication methods via the environment:

- [Spacelift API tokens](#spacelift-api-tokens).
- [GitHub tokens](#github-tokens).
- [Spacelift API keys](#spacelift-api-keys).

The Spacelift CLI looks for authentication configurations in the order specified above,
and will stop as soon as it finds a valid configuration. For example, if a Spacelift API
token is specified, GitHub tokens and Spacelift API keys will be ignored, even if their
environment variables are specified.

#### Spacelift API tokens

Spacelift API tokens can be specified using the `SPACELIFT_API_TOKEN` environment variable.
When this variable is found, the CLI ignores all the other authentication environment variables
because the token contains all the information needed to authenticate.

NOTE: API tokens are generally short-lived and will need to be re-created often.

#### GitHub tokens

GitHub tokens are only available to accounts that use GitHub as their identity provider, but
are very convenient for use in GitHub actions. To use a GitHub token, set the following
environment variables:

- `SPACELIFT_API_ENDPOINT` - the URL to your Spacelift account, for example `https://mycorp.app.spacelift.io`.
- `SPACELIFT_API_GITHUB_TOKEN` - a GitHub personal access token.

#### Spacelift API keys

To use a Spacelift API key, set the following environment variables:

- `SPACELIFT_API_ENDPOINT` - the URL to your Spacelift account, for example `https://mycorp.app.spacelift.io`.
- `SPACELIFT_API_KEY_ID` - the ID of your Spacelift API key. Available via the Spacelift application.
- `SPACELIFT_API_KEY_SECRET` - the secret for your API key. Only available when the secret is created.

More information about API authentication can be found at <https://docs.spacelift.io/integrations/api#authenticating-with-the-api>.
```

adamconnelly (commented):
Looks great - sorry if some of my suggestions are quite picky - you can let me know if I'm going over the top 😆

README.md
marcinwyszynski: I understand what you mean but your version does not allow the reader to understand the order of precedence.

README.md
```diff
@@ -1,2 +1,188 @@
...
+
+When certain variables are defined in the environment, `spacelift-cli` will try to use them as Spacelift API credentials. First, the `SPACELIFT_API_TOKEN`. IF this variable is present in the environment, it alone contains enough information to set up a Spacelift session. Hence, the CLI will look no further and attempt to use it to talk to the server. Note that these tokens are generally short-lived.
+
+If the `SPACELIFT_API_TOKEN` variable is not present, the CLI will look for the
+`SPACELIFT_API_ENDPOINT` variable. This one contains the API endpoint to talk to, essentially pointing to your account and (optionally) Spacelift server - for example, `https://mycorp.app.spacelift.io`. If this variable is not present, environment lookup fails. Otherwise, it proceeds to credentials lookup.
```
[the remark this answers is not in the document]
marcinwyszynski: Sure. It means a different API server. All the vanity subdomains in `app.spacelift.io` end up hitting the same load balancer, but our preprod is in `app.spacelift.dev` and if/once we spin up the US region, the API server for that may be `app.spacelift.us`. Similar for on-prems, if we ever have them. So this is pure future-proofing.

README.md
```diff
@@ -1,2 +1,188 @@
...
+NAME:
+   spacelift-cli account - Manage Spacelift accounts
+
+USAGE:
+   spacelift-cli account command [command options] [arguments...]
+
+COMMANDS:
+   login    Log in to a Spacelift account
+   logout   Log out of an existing Spacelift account
+   select   Select one of existing Spacelift accounts
```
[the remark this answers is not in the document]
marcinwyszynski: Probably, as a separate change. Also, `list`.

README.md
```diff
@@ -1,2 +1,188 @@
...
+   --id value  User-facing ID (slug) of the stack
+   --help, -h  show help (default: false)
+```
+
+Each of these operations will require user-facing stack ID (so called "slug") visible - among other places - in the stack URL in the GUI. Let's look at the operations one by one.
+
+#### `stack deploy` subcommand
+
+```bash
+❯ spacelift-cli stack --id=stack-id deploy --help
```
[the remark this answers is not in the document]
marcinwyszynski: Unfortunately, it is :(

README.md
marcinwyszynski: It would. I like the word `deployment` better but historically we called them `tracked runs`, and that's passed to policies which folks may have already written, so I'm trying to steer us gently towards what I see as a the better (more universal) name.

You know, naming things is hard, especially in a foreign language.

README.md
marcinwyszynski: Code reuse :) I guess you're right, this sucks.

One way around it would be for the logs subroutine to return a tuple of `(*RunStateTransition, error)` and have the caller decide what to do about it. I'll try to fix it in this iteration.

README.md
adamconnelly: Yeah - I tried to cover that with the following, but it's maybe not super clear:

> The Spacelift CLI looks for authentication configurations in the order specified above,
> and will stop as soon as it finds a valid configuration.

Anyhow - like I say it's really subjective 🤷

adamconnelly (approved):
Looking good! 🚀

README.md
marcinwyszynski: I missed that entirely.

README.md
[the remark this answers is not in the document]
adamconnelly: Cool - that makes sense. I can't quite make my mind up here, but it might be worth just removing the `(optionally)` on the grounds that it's not optional (i.e. you always need to provide the `app.spacelift.io` part).

README.md
adamconnelly: Yeah, and deprecating things is really tough as well, so I'm definitely not criticising here - just wondered.

README.md
marcinwyszynski: Done.

README.md
marcinwyszynski: Accepted the suggestion.

PR #3, opened 2021-04-08
Add current account and a few other minor tweaks

adamconnelly:
## `account current` command

Added a new command to output the current selected account:

```shell
$ go run spacelift-cli.go -- account current
adamconnelly-admin
```

Here's the usage:

```shell
$ go run spacelift-cli.go -- account current --help
NAME:
   spacelift-cli account current - Outputs the account you currently have selected

USAGE:
   spacelift-cli account current [command options]

OPTIONS:
   --help, -h  show help (default: false)
```

## Adding missing newlines

I noticed while logging in my shell was displaying a `%` after the commands completed. For example:

```shell
$ go run spacelift-cli.go -- account login adamconnelly-admin
Enter Spacelift endpoint (eg. https://unicorn.app.spacelift.io/): https://adamconnelly.app.spacelift.io/
Select credentials type: 1 for API key, 2 for GitHub access token: 1
Enter API key ID: ***
Enter API key secret: %
$
```

It now looks like this:

```shell
$ go run spacelift-cli.go -- account login adamconnelly-admin
Enter Spacelift endpoint (eg. https://unicorn.app.spacelift.io/): https://adamconnelly.app.spacelift.io
Select credentials type: 1 for API key, 2 for GitHub access token: 1
Enter API key ID: ***
Enter API key secret:
$
```

## ArgsUsage

I've updated the account commands that accept an alias so that's reflected in the usage. Before it just displayed a generic `[arguments...]` text:

```text
USAGE:
   spacelift-cli account logout [command options] [arguments...]
```

Now it looks like this:

```shell
USAGE:
   spacelift-cli account logout [command options] <account-alias>
```

### DIFF
--- a/cmd/internal/profile/profile.go
+++ b/cmd/internal/profile/profile.go
@@ -36,6 +36,7 @@ func Command() *cli.Command {
 			return nil
 		},
 		Subcommands: []*cli.Command{
+			currentCommand(),
 			loginCommand(),
 			logoutCommand(),
 			selectCommand(),
--- a/cmd/internal/profile/current_command.go
+++ b/cmd/internal/profile/current_command.go
@@ -12,6 +12,10 @@ func currentCommand() *cli.Command {
 	return &cli.Command{
 		Name:  "current",
 		Usage: "Outputs the account you currently have selected",
+
+		// Use a space to cause the args usage to not be displayed since the `current` command
+		// doesn't accept any arguments
+		ArgsUsage: " ",
 		Action: func(ctx *cli.Context) error {
 			if _, err := os.Lstat(currentPath); err != nil {
 				return fmt.Errorf("no account is currently selected: %w", err)
--- a/cmd/internal/profile/current_command.go
+++ b/cmd/internal/profile/current_command.go
@@ -1,0 +1,31 @@
+package profile
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/urfave/cli/v2"
+)
+
+func currentCommand() *cli.Command {
+	return &cli.Command{
+		Name:  "current",
+		Usage: "Outputs the account you currently have selected",
+		Action: func(ctx *cli.Context) error {
+			if _, err := os.Lstat(currentPath); err != nil {
+				return fmt.Errorf("no account is currently selected: %w", err)
+			}
+
+			linkTarget, err := os.Readlink(currentPath)
+			if err != nil {
+				return fmt.Errorf("could not find the target of the current account symlink: %w", err)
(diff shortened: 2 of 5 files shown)


cmd/internal/account/before_each.go
```diff
@@ -18,11 +18,20 @@ var (
 )
 
 func beforeEach(cliCtx *cli.Context) error {
-	if nArgs := cliCtx.NArg(); nArgs != 2 {
-		return fmt.Errorf("expecting account alias as the only argument, got %d instead", nArgs)
+	var nArgs int
```
adamconnelly, line 21: I kinda wonder if it would make sense to allow each subcommand to specify their required arguments, but for now I just went with this approach since I'm still a Go noob, so I didn't want to complicate things.

cmd/internal/account/current_command.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+func currentCommand() *cli.Command {
+	return &cli.Command{
+		Name:  "current",
+		Usage: "Outputs the account you currently have selected",
+
+		// Use a space to cause the args usage to not be displayed since the `current` command
+		// doesn't accept any arguments
+		ArgsUsage: " ",
```
adamconnelly, line 18: This feels a bit hacky. I doubt it would cause any issues, but I'm happy to change it / do more investigation if folk aren't happy with it.

cmd/internal/account/before_each.go
marcinwyszynski: Yes, I think that's the way to go. I'll fix this in my PR instead. I just tried to be too clever there.

cmd/internal/account/current_command.go
```diff
@@ -0,0 +1,35 @@
...
+				return fmt.Errorf("no account is currently selected: %w", err)
+			}
+
+			linkTarget, err := os.Readlink(currentPath)
+			if err != nil {
+				return fmt.Errorf("could not find the target of the current account symlink: %w", err)
+			}
+
+			alias := filepath.Base(linkTarget)
+			fmt.Printf("%s\n", alias)
```
marcinwyszynski, line 30: Nit: `fmt.Println`

marcinwyszynski (approved):

cmd/internal/profile/profile.go
```diff
@@ -39,6 +39,7 @@ func Command() *cli.Command {
 			loginCommand(),
 			logoutCommand(),
 			selectCommand(),
+			currentCommand(),
```
marcinwyszynski, line 42: Nit: alphasort.

PR #5, opened 2021-04-09
Simplify binary names inside archives

adamconnelly:
- Added a tar transform that changes the architecture specific binary names (e.g. `spacectl-darwin-amd64`) to `spacectl` since the archive contains the architecture in the name. This means that you can just untar without having to rename the file afterwards.
- Used a specific transform for Windows that adds the `.exe` extension.


marcinwyszynski (approved):
❤️

But before your merge, can you please update the README to reflect this change?

cube2222 (approved):

PR #6, opened 2021-04-12
Pull profile code out to a ProfileManager

adamconnelly:
- Created a profile manager and added unit tests for profile management.
- Updated spacectl to use the manager.


client/session/profile_manager.go
```diff
@@ -0,0 +1,242 @@
...
+// NewProfileManager creates a new ProfileManager using the specified directory to store the profile data.
+func NewProfileManager(profilesDirectory string) *ProfileManager {
+	return &ProfileManager{
+		ProfilesDirectory: profilesDirectory,
+		CurrentPath:       filepath.Join(profilesDirectory, CurrentFileName),
+	}
+}
+
+// Init initialises the profile manager, making sure it is ready to store and retrieve profiles.
+func (m *ProfileManager) Init() error {
```
adamconnelly, line 51: Would it be more normal to include this logic in `NewProfileManager`?

marcinwyszynski: I'm not an ideologue, so no opinion here.
@cube2222 any thoughts?

client/session/profile_manager_test.go
```diff
@@ -0,0 +1,341 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"testing"
+
+	"github.com/franela/goblin"
+	. "github.com/onsi/gomega"
+	"github.com/spacelift-io/spacectl/client/session"
```
marcinwyszynski, line 12: Nit: import grouping. At Spacelift, our practice is to split imports into up to 3 groups:

- stdlib imports;
- 3rd party imports;
- local imports;

client/session/profile_manager_test.go
```diff
@@ -0,0 +1,341 @@
...
+			}
+
+			profilesDirectory = path.Join(testDirectory, "profiles")
+
+			manager = session.NewProfileManager(profilesDirectory)
+		})
+
+		g.AfterEach(func() {
+			if err := os.RemoveAll(testDirectory); err != nil {
+				g.Fail(fmt.Errorf("Failed to cleanup temp profiles directory: %w", err))
```
marcinwyszynski, line 37: I'm not 100% sure you can fail in before/after hooks in Goblin. I used `testify/require` in those cases.

client/session/profile_manager_test.go
```diff
@@ -0,0 +1,341 @@
...
+			}
+		})
+
+		g.Describe("Init", func() {
+			g.Describe("profiles directory doesn't exist", func() {
+				g.It("creates directory", func() {
+					manager.Init()
+
+					if _, err := os.Stat(profilesDirectory); err != nil {
```
marcinwyszynski, line 46: Why not use Gomega assertions here?

client/session/profile_manager_test.go
adamconnelly: Before I go looking - do you guys just manually do that, or do you have your editors configured to do that automatically / can we lint it?

client/session/profile_manager_test.go
```diff
@@ -0,0 +1,341 @@
...
+						testProfile := &session.Profile{
+							Alias:       profileAlias,
+							Credentials: createValidGitHubCredentials(),
+						}
+
+						err := manager.Create(testProfile)
+						Expect(err).To(BeNil())
+
+						savedProfile, err := manager.Get(testProfile.Alias)
+						if err != nil {
```
marcinwyszynski, line 117: Use Gomega assertion here.

marcinwyszynski (commented):
The PR looks good, but the test is not very idiomatic, and sometimes feels wrong, too (fails inside setup/teardown lambdas).

client/session/profile_manager_test.go
marcinwyszynski: I do it manually, but once I have 3 groups, `goimports` somehow "knows" where to put each new import. Maybe there's a way to automate that, haven't looked into it at all.

client/session/profile_manager_test.go
adamconnelly: @marcinwyszynski I've just tried it out there, and it looks like using `Fail()` works, but using `require` causes the test run to hang.

For example, I tried adding the following to a `BeforeEach()`:

```go
require.Nil(t, "abc123", "Testing")
```

If I do that, the test execution just seems to pause at that point. But if I do the following it fails the tests correctly:

```go
g.Fail("testing 123")
```

marcinwyszynski: OK, my bad then.

adamconnelly:
@marcinwyszynski I've made the changes to sort the imports and switch to Gomega assertions for a few places that I could.

I haven't updated the fails inside setups/teardowns yet because they actually worked when I tried them just now, but using `require` didn't. I might be doing something weird there though.

RE the tests not being very idiomatic - were there other issues with them other than the things that you mentioned? If so I'm happy to rework them - just need some pointers.

marcinwyszynski (approved):

marcinwyszynski:
All good. Re: idiomatic approach, I mostly meant not using Gomega assertions.

client/session/profile_manager.go
michieldewilde: I would move this to the `NewProfileManager`. 
A one line function for setting up the the manager wouldn't need a separate function.

cube2222: No strong opinions. I guess a separate Init is a non-formal contract the user of the ProfileManager must uphold, so one more thing to think about, whereas by moving it into NewProfileManager there's less room for misuse.

michieldewilde: We could even use `init()` for this 😄

client/session/profile_manager.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	if profile.Credentials.KeySecret == "" {
+		return errors.New("'KeySecret' must be provided for API Key credentials")
+	}
+
+	return nil
+}
+
+func setCurrentProfile(profilePath string, manager *ProfileManager) error {
```
michieldewilde, line 196: passing `*ProfileManager` is counter intuitive.
It would make more sense to make this a method on the `ProfileManager`.
```
func (pm *ProfileManageer) setCurrentProfile
```
Same for `func writeProfileToFile(profile *Profile, manager *ProfileManager)`

cube2222 (approved):
LGTM!

client/session/profile_manager.go
cube2222: **no**

client/session/profile_manager.go
adamconnelly: Cheers - I honestly didn't even consider that!

client/session/profile_manager.go
adamconnelly: It was actually worth doing - I realised when I merged them that I had a place where it wasn't being initialised 🤦

adamconnelly:
Thanks for all the feedback everyone - after thinking about it I've reworked the PR to be backwards compatible with the existing config storage. I just figured on the off-chance that someone's downloaded and tried out spacectl, I didn't want to immediately break things for them, especially when it wasn't a big deal to avoid breaking.

I've also made changes for the other feedback.

PR #7, opened 2021-04-13
Add `profile list` command

adamconnelly:
Added a new command that lists the stored profiles. At the moment it just output them to the console in a table for simplicity.

The profiles are output in alphabetical order, with the current profile indicated via a `*`.

Here's what the output looks like:

```shell
$ go run . -- profile list
Current | Alias               | Endpoint                              | Type
*       | adamconnelly-apikey | https://adamconnelly.app.spacelift.io | API Key
        | adamconnelly-github | https://adamconnelly.app.spacelift.io | GitHub
```

Here's what it looks like after switching to my other profile:

```shell
Current | Alias               | Endpoint                              | Type
        | adamconnelly-apikey | https://adamconnelly.app.spacelift.io | API Key
*       | adamconnelly-github | https://adamconnelly.app.spacelift.io | GitHub
```

Here's what it looks like if no profiles are stored:

```shell
$ go run . -- profile list
Current | Alias | Endpoint | Type
```

Here's the command usage info:

```shell
NAME:
   spacectl profile list - List all your Spacelift account profiles

USAGE:
   spacectl profile list [command options]

OPTIONS:
   --help, -h  show help (default: false)
```


client/session/profile_manager_test.go
```diff
@@ -314,6 +315,33 @@ func TestProfileManager(t *testing.T) {
...
+
+				// Sort the slice to guarantee the order when comparing the results
+				sort.SliceStable(profiles, func(i int, j int) bool {
+					return profiles[i].Alias < profiles[j].Alias
+				})
+
+				Expect(len(profiles)).To(Equal(3))
+				Expect(profiles[0].Alias).To(Equal("profile-1"))
+				Expect(profiles[1].Alias).To(Equal("profile-2"))
+				Expect(profiles[2].Alias).To(Equal("profile-3"))
```
adamconnelly, line 342: It feels like there's probably a better way of doing this, but I couldn't spot anything obvious in the Gomega assertions. Any ideas / tips?

cube2222: Actually, I'm quite sure that's the way to go

client/session/stored_credentials.go
```diff
@@ -23,6 +23,11 @@ const (
 	CredentialsTypeGitHubToken
 )
 
+// String returns the string representation of the type.
+func (t CredentialsType) String() string {
+	return [...]string{"Invalid", "API Key", "GitHub"}[t]
```
cube2222, line 28: Today I learned what Array syntax is 😯
```
[...]string{
```
Though, do I understand correctly that []string{...} would be perfectly ok here and this just omits one allocation?

cube2222 (approved):

client/session/stored_credentials.go
adamconnelly: Well this is slightly awkward... the truth is that I just copied it from an example 😆 I was trying to search for something like "what's best practice in Go for converting enum values to strings" and found this page: https://yourbasic.org/golang/iota/

Now you've put me on the spot I think you're right. `[...]string{"1", "2"}` allocates an array with a length of two, and `[]string{"1", "2"}` allocates an array with a length of two, and wraps it in a slice from my quick research. The `...` syntax is just used to tell the compiler to automatically set the array length, so `[...]string{"1", "2"}` is equivalent to `[2]string{"1", "2"}`.

My guess is the reasons for doing it like this are:

- To avoid having to update the array length as new items are added.
- To avoid creating a slice when all we're ever doing with the array is directly accessing an element by its index.

I'm guessing that using a slice here would be totally fine as well (in the sense that it's unlikely to really make a different to performance for this application).

Issue #8, opened 2021-04-16
Spacectl current profile doesn't work on Windows


adamconnelly:
Setting the current profile doesn't work on Windows because it relies on symlinks. If you try to set the current profile, for example, you get an error like the following:

```shell
> spacectl.exe profile select adamconnelly
2021/04/16 09:19:40 could not symlink the config file for adamconnelly: symlink C:\Users\adam\.spacelift\adamconnelly C:\Users\adam\.spacelift\current: A required privilege is not held by the client.
```

It might be simpler for now to switch to a single json file that has a property for the current profile.

adamconnelly:
I've altered the config format to just store all the profile data in a single config file instead of using symlinks to point at the current profile. This should work on all OSes.

PR #9, opened 2021-04-16
Run unit tests on GitHub Actions

adamconnelly:
- Added a new workflow to run unit tests. I've configured it to run on Linux, Windows and MacOS so we can have confidence the tool works on all of them.
- I've added some extra assertions to the profile tests to give us better output when they fail. Previously it was just segfaulting on Windows because of #8.

### DIFF
--- a/.github/workflows/unit-testing.yml
+++ b/.github/workflows/unit-testing.yml
@@ -1,0 +1,30 @@
+name: Unit testing
+
+on:
+  push: { branches: [main] }
+  pull_request: { branches: [main] }
+
+jobs:
+  unit-testing:
+    name: Test the code
+    strategy:
+      matrix:
+        os:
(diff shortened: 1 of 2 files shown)


adamconnelly:
I've setup this pipeline to run on Linux, Windows and MacOS because I was hoping to catch the issue in #8 where symlinks aren't working on Windows. The issue is that the bug only happens if you're running as a non-elevated account, and GitHub Actions run elevated.

Unfortunately I haven't been able to figure out a way to "de-escalate" the account, so at this stage I'm not sure how worthwhile it is running the tests on all OSes if it's not going to catch problems like this. OTOH it probably doesn't hurt 🤷

marcinwyszynski (approved):

b4k3r (approved):

PR #11, opened 2021-04-28
Add local preview.

cube2222:


adamconnelly (approved):
Neat - that looks really cool!

One minor comment about the structure is do you think it's worth creating specific functions in the GraphQL client to take some of the implementation details of making the GraphQL requests out of the command implementation?

I'm just thinking that since we've effectively created a client-library for accessing the API, what about defining the structs in the client library rather than inline and being able to do `client.CreateUploadRequest()`, `client.RunLocalChanges()`, etc?

cube2222:
> Neat - that looks really cool!
> 
> One minor comment about the structure is do you think it's worth creating specific functions in the GraphQL client to take some of the implementation details of making the GraphQL requests out of the command implementation?
> 
> I'm just thinking that since we've effectively created a client-library for accessing the API, what about defining the structs in the client library rather than inline and being able to do `client.CreateUploadRequest()`, `client.RunLocalChanges()`, etc?

@adamconnelly I'm against it TBH. GraphQL is already very easy to use ad-hoc, and the functions would probably only be used in a single place. If something is used in a lot of places, sure. Like the runLogs helper.

internal/cmd/stack/flags.go
```diff
@@ -30,3 +30,9 @@ var flagTail = &cli.BoolFlag{
 	Usage: "Indicate whether to tail the run",
 	Value: false,
 }
+
+var flagTailDefaultTrue = &cli.BoolFlag{
```
adamconnelly, line 34: One more thing actually - having a flag default to true feels a little strange for a CLI because you don't typically use them with an argument. For example you'd normally do:

```shell
spacectl some-command
```

or

```shell
spacectl some-command -tail
```

But you wouldn't usually do

```shell
spacectl some-command -tail false
```

Might just be worth having a think about it. Maybe flipping it and having an `-interactive` flag or something like that? That still has issues though because I get that you want to default to tailing in this case. 🤔 

Thinking about this now - are we using the right term here at all? I'd normally think of `tail` as "return the last X records", and `follow` as "keep returning results as they're available".

cube2222: Maybe I'll just add -notail ?

adamconnelly: Yeah - that could work. It means there's not two separate defaults for `-tail` depending on the command used as well.

cube2222: Done.

PR #12, opened 2021-05-04
Use goreleaser, add version, update JWT spec

marcinwyszynski:
Few changes:

- use [goreleaser](https://goreleaser.com/) for releasing the binary;
- add top-level `version` command;
- update the JWT library to handle tokens with array `aud` field;


.github/workflows/release.yml
```diff
@@ -1,36 +1,39 @@
 name: Release
 
-on:
-  push: { tags: ["v*"] }
+on: { push }
```
adamconnelly, line 3: Do we definitely want to run this for every push rather than just for tags?

adamconnelly (approved):
LGTM - just one small question about the branch trigger.

.github/workflows/release.yml
marcinwyszynski: Note line 33.

PR #14, opened 2021-05-17
Add run confirmation and user provided metadata.

cube2222:


client/client.go
```diff
@@ -19,7 +20,7 @@ type client struct {
 
 // New returns a new instance of a Spacelift Client.
 func New(wraps *http.Client, session session.Session) Client {
```
marcinwyszynski, line 22: Could we perhaps reuse my fork of the GraphQL client that we're already using in our Terraform provider?
https://github.com/marcinwyszynski/graphql

marcinwyszynski (changes requested):

client/client.go
cube2222: Sure

marcinwyszynski (approved):

Issue #17, opened 2021-05-27
spacectl download link in README is broken


jonassteinberg1:
```
wget https://github.com/spacelift-io/spacectl/releases/latest/download/spacectl-darwin-amd64.tar.gz
--2021-05-27 14:56:23--  https://github.com/spacelift-io/spacectl/releases/latest/download/spacectl-darwin-amd64.tar.gz
Resolving github.com (github.com)... 140.82.113.3
Connecting to github.com (github.com)|140.82.113.3|:443... connected.
HTTP request sent, awaiting response... 302 Found
Location: https://github.com/spacelift-io/spacectl/releases/download/v0.6.0/spacectl-darwin-amd64.tar.gz [following]
--2021-05-27 14:56:24--  https://github.com/spacelift-io/spacectl/releases/download/v0.6.0/spacectl-darwin-amd64.tar.gz
Reusing existing connection to github.com:443.
HTTP request sent, awaiting response... 404 Not Found
2021-05-27 14:56:24 ERROR 404: Not Found.
```

adamconnelly:
@jonassteinberg1 thanks a lot for letting us know about this!

@marcinwyszynski I think the issue must be that we've changed the filename when switching to GoReleaser. It looks like we can use a URL like https://github.com/spacelift-io/spacectl/releases/latest/download/spacectl_0.6.0_linux_amd64.zip to get the latest version. The issue is that we've got the version number in the filename, so we can't generate a static URL using that.

Is it safe to just drop the version number from the zip filename? If so I'll sort it and update the README.

marcinwyszynski:
Yup, the migration to GoReleaser is likely what caused it. Not sure about the URL change TBH, need to verify that with some of our clients who are downloading it programmatically.

adamconnelly:
I'm just closing this issue for now since the README has been updated.

PR #18, opened 2021-06-02
Remove curl instructions from README

adamconnelly:
Until we can come up with a static URL that can be used to download spacectl, just point users to the releases page.

Issue: #17


marcinwyszynski (approved):
Fair enough.

PR #44, opened 2021-09-07
Set/delete environment

b4k3r:


marcinwyszynski (changes requested):
I don't like the API. It should be more clear what you're adding. I'd expect two separate subcommands like `setenv` and `mount`. The latter should support pointing to an existing file (via path) or reading from stdin (eg. piping output of some commant) and handle things like reading and base64 encoding.

marcinwyszynski (approved):

internal/cmd/stack/environment.go
```diff
@@ -0,0 +1,157 @@
...
+			WriteOnly bool   `graphql:"writeOnly"`
+			Value     string `graphql:"value"`
+		} `graphql:"stackConfigAdd(stack: $stack, config: $config)"`
+	}
+
+	variables := map[string]interface{}{
+		"stack": graphql.ID(stackID),
+		"config": ConfigInput{
+			ID:        graphql.ID(envName),
+			Type:      ConfigType("ENVIRONMENT_VARIABLE"),
```
adamconnelly, line 51: Maybe define these as constants next to the type? Just since you've created a custom type for it anyway.

internal/cmd/stack/environment.go
```diff
@@ -0,0 +1,157 @@
...
+			Value:     graphql.String(envValue),
+			WriteOnly: graphql.Boolean(cliCtx.Bool(flagEnvironmentWriteOnly.Name)),
+		},
+	}
+
+	if err := authenticated.Client.Mutate(context.Background(), &mutation, variables); err != nil {
```
adamconnelly, line 57: You can use `cliCtx.Context` here. Not sure if it would make any difference, but figure if there's already a context available we might as well use it.

internal/cmd/stack/environment.go
```diff
@@ -0,0 +1,157 @@
...
+			ID        string     `graphql:"id"`
+			WriteOnly bool       `graphql:"writeOnly"`
+			Type      ConfigType `graphql:"type"`
+		} `graphql:"stackConfigAdd(stack: $stack, config: $config)"`
+	}
+
+	variables := map[string]interface{}{
+		"stack": graphql.ID(stackID),
+		"config": ConfigInput{
+			ID:        graphql.ID(envName),
```
adamconnelly, line 109: Should we do any kind of validation to make sure this is a valid unix path?

internal/cmd/stack/environment.go
```diff
@@ -0,0 +1,157 @@
...
+type ConfigInput struct {
+	ID        graphql.ID      `json:"id"`
+	Type      ConfigType      `json:"type"`
+	Value     graphql.String  `json:"value"`
+	WriteOnly graphql.Boolean `json:"writeOnly"`
+}
+
+func setVar(cliCtx *cli.Context) error {
+	if nArgs := cliCtx.NArg(); nArgs != 2 {
+		return fmt.Errorf("expecting environment setenv as two arguments, got %d instead", nArgs)
```
adamconnelly, line 31: I'd maybe reword this to

```
expected two arguments to `environment setenv` but got %d
```

internal/cmd/stack/environment.go
```diff
@@ -0,0 +1,157 @@
...
+			return fmt.Errorf("couldn't read from STDIN: %w", err)
+		}
+	case 2:
+		filePath := cliCtx.Args().Get(1)
+
+		if fileContent, err = os.ReadFile(filePath); err != nil {
+			return fmt.Errorf("couldn't read file from %s: %w", filePath, err)
+		}
+	default:
+		return fmt.Errorf("expecting environment mount as max two arguments, got %d instead", nArgs)
```
adamconnelly, line 95: Same here - I'd reword this slightly.

internal/cmd/stack/environment.go
b4k3r: We do this on the server side.

adamconnelly (approved):

PR #50, opened 2022-02-17
Login with a web browser

b4k3r:


adamconnelly (approved):
LGTM - that's awesome!!!

internal/cmd/profile/get_alias_with_profile.go
```diff
@@ -0,0 +1,40 @@
...
+	manager, err := session.NewProfileManager(filepath.Join(userHomeDir, session.SpaceliftConfigDirectory))
+	if err != nil {
+		return fmt.Errorf("could not create profile manager: %w", err)
+	}
+
+	profile := manager.Current()
+	if profile != nil && profile.Credentials.Type == session.CredentialsTypeAPIToken {
+		apiTokenProfile = profile
+	} else {
+		return errors.New("command is supported only for exisiting API Token profile. Please use `spacectl profile login <alias>` instead")
```
adamconnelly, line 36:
```suggestion
		return errors.New("command is only supported when using an existing API Token profile. Please use `spacectl profile login <alias>` instead")
```

There was a typo in `exisiting`, but I've made a suggestion about rewording slightly. Mainly `s/supported only/only supported` - both are valid, but IMO "only supported" reads more naturally.

adamconnelly (approved):
Awesome!!! :rocket:

wojciech12 (approved):

PR #61, opened 2022-05-15
Change version command into a flag

spitzzz:
# What's in this PR
* Removed the `version` command and switched to using `--version` instead (as a root flag)


main.go
```diff
@@ -3,24 +3,25 @@ package main
...
 
 var version = "dev"
 
 func main() {
 	app := &cli.App{
-		Name:  "spacectl",
-		Usage: "Programmatic access to Spacelift GraphQL API",
+		Name:     "spacectl",
+		Version:  version,
+		Compiled: time.Now(),
```
adamconnelly, line 20: I don't think this is correct - it'll return the current time rather than the compilation time. I think what we'd need to do is something similar to the `version` variable - create a var for the compilation time and then set it via the linker.

It looks like goreleaser supports this out the box: https://goreleaser.com/customization/build/. Based on the information about the `.Date` variable [here](https://goreleaser.com/customization/templates/), I think what you'd need to do is:

- Declare a var called `date` as a string.
- Parse it using [`time.Parse()`](https://pkg.go.dev/time#Parse).

adamconnelly (commented):
@spitzzz I prefer the `--version` flag as well - the only thing is that this is technically a backwards-incompatible change. I think there's probably little risk in removing it, but just wanted to point that out.

main.go
adamconnelly: Yeah - in a way it's actually safer to do what you've done and just re-add the version command as-is, without changing its format. That way we don't worry about it, and we can either make it consistent or drop the command completely at some point in the future.

adamconnelly (approved):
I left a couple more really minor comments, but I'm approving the PR. Just sort them and then feel free to merge.

Thanks :bow:

Issue #33, opened 2021-06-25
Add missing stack commands


adamconnelly:
We already have a number of stack commands, but some functionality is missing. Go through https://github.com/spacelift-io/spacectl/blob/main/specs/spacectl-to-v1.md#stacks, add any missing commands, and adjust existing commands if necessary.

Also, maybe the `spacectl stack show` command would accomplish what's been asked for in https://github.com/spacelift-io/spacectl/issues/20?

Insomniak47:
A comment on a note from the doc
> Do we want to have any commands for getting stack resources, or is that overkill?

I think having a terraform show equivalent would be very useful for common workflows

PR #66, opened 2022-05-27
Add ability to discard unconfirmed run

jmfontaine:
Added a new `spacectl stack discard` command to discard unconfirmed runs.

Issues: #33

### DIFF
--- a/internal/cmd/stack/run_discard.go
+++ b/internal/cmd/stack/run_discard.go
@@ -27,18 +27,13 @@ func runDiscard() cli.ActionFunc {
 
 		ctx := context.Background()
 
-		var requestOpts []graphql.RequestOption
-		if cliCtx.IsSet(flagRunMetadata.Name) {
-			requestOpts = append(requestOpts, graphql.WithHeader(UserProvidedRunMetadataHeader, cliCtx.String(flagRunMetadata.Name)))
-		}
-
-		if err := authenticated.Client.Mutate(ctx, &mutation, variables, requestOpts...); err != nil {
+		if err := authenticated.Client.Mutate(ctx, &mutation, variables); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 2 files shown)


internal/cmd/stack/run_discard.go
```diff
@@ -0,0 +1,58 @@
...
+
+		variables := map[string]interface{}{
+			"stack": graphql.ID(stackID),
+			"run":   graphql.ID(cliCtx.String(flagRun.Name)),
+		}
+
+		ctx := context.Background()
+
+		var requestOpts []graphql.RequestOption
+		if cliCtx.IsSet(flagRunMetadata.Name) {
```
marcinwyszynski, line 31: I think this flag makes no sense when discarding the run.

internal/cmd/stack/run_discard.go
```diff
@@ -0,0 +1,58 @@
...
+			requestOpts = append(requestOpts, graphql.WithHeader(UserProvidedRunMetadataHeader, cliCtx.String(flagRunMetadata.Name)))
+		}
+
+		if err := authenticated.Client.Mutate(ctx, &mutation, variables, requestOpts...); err != nil {
+			return err
+		}
+
+		fmt.Println("You have successfully discarded a deployment")
+
+		fmt.Println("The live run can be visited at", authenticated.Client.URL(
```
marcinwyszynski, line 41: The copy is misleading, at this stage it's no longer a "live" run.

internal/cmd/stack/run_discard.go
jmfontaine: Agreed.

internal/cmd/stack/run_discard.go
jmfontaine: Good catch!

marcinwyszynski (approved):

PR #70, opened 2022-08-04
feat(spacectl/stack): add stack output command

garry-harthill-cko:
Add a command to the stack menu which displays the current outputs
of the stack.

### DIFF
--- a/internal/cmd/stack/stack.go
+++ b/internal/cmd/stack/stack.go
@@ -176,7 +176,7 @@ func Command() *cli.Command {
 			},
 			{
 				Name:  "outputs",
-				Usage: "Shows current outputs for a specific stack. Does not show sensitive outputs.",
+				Usage: "Shows current outputs for a specific stack. Does not show the value of sensitive outputs.",
 				Flags: []cli.Flag{
 					flagStackID,
 					flagOutputID,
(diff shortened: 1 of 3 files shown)


jmfontaine:
Thanks for the PR @garry-harthill-cko! I will have the Engineering team review it.

internal/cmd/stack/stack.go
```diff
@@ -174,6 +174,19 @@ func Command() *cli.Command {
 					},
 				},
 			},
+			{
+				Name:  "outputs",
+				Usage: "Shows current outputs for a specific stack. Does not show sensitive outputs.",
```
adamconnelly, line 179: Maybe we need to tweak this to "Does not show the value of sensitive outputs", on the grounds you're displaying them but just masking the value?

adamconnelly (commented):
@garry-harthill-cko thanks a lot for the contribution - it looks great! I've just left a small suggestion about the usage message for the command. Let me know what you think.

garry-harthill-cko:
> @garry-harthill-cko thanks a lot for the contribution - it looks great! I've just left a small suggestion about the usage message for the command. Let me know what you think.

Good point. It does still display the sensitive outputs. Just not the values.

adamconnelly (approved):
LGTM - thanks again Garry :bow:

PR #72, opened 2022-08-11
feat(spacectl/stack): add environment list command

jpJuni0r:


internal/cmd/stack/stack.go
```diff
@@ -151,6 +151,16 @@ func Command() *cli.Command {
 						Before:    authenticated.Ensure,
 						ArgsUsage: "NAME VALUE",
 					},
+					{
+						Name:  "getvar",
```
marcinwyszynski, line 155: Shouldn't it be `getenv`?

adamconnelly:
@jpJuni0r - thanks a lot for the contribution! Overall it looks good, but I've got a few suggestions.

## Naming

I know you and Marcin have had a discussion about naming. I'd like to suggest calling the new command `list`, for example `spacectl stack environment list`. That's then consistent with `spacectl stack list`.

I realise we've got a command called `setvar`, but that's because we support both environment variables and mounted files, and there's a separate `mount` command. So this way we'd have:

- `list` - lists all the env vars and mounted files for a stack.
- `setvar` - sets an environment variable.
- `mount` - mounts a file.

## Contexts

We support both stack-level environment variables, as well as those defined in a [context](https://docs.spacelift.io/concepts/configuration/context.html) and attached to a stack, so I think this command should handle both for consistency with the UI.

Here's the GraphQL query the frontend uses for reference:

```gql
query GetEnvironment($stackId: ID!) {
  stack(id: $stackId) {
    id
    
    # I actually don't think you need `attachedContexts` for this - I've just included it for reference in case I'm missing something
    attachedContexts {
      id
      contextName
      config {
        ...configElement
      }
    }

    # I think all the info we need is in the runtimeConfig property, including the context the config item came from
    runtimeConfig {
      context {
        contextId
        contextName
      }
      element {
        ...configElement
      }
    }
  }
}
fragment configElement on ConfigElement {
  createdAt
  id
  runtime
  type
  value
  writeOnly
  fileMode
}
```

## Output

At the moment you're just displaying the name and value, but I think it would be useful to add some more info. Maybe something like this:

| Name | Type | Value | File Mode | Context |
| ---- | ---- | ----- | --------- | ------- |
| MY_VAR | Variable | abc123 | | |
| MY_SECRET_VAR | Variable | ***** | | |
| /mnt/workspace/some-file | File | | 755 | |
| MY_CONTEXT_VAR | Variable | def321 | | My Context |
| AWS_ACCESS_KEY_ID | Variable | `<computed>` | | |

Maybe if that becomes too messy we could separate env vars and mounted files into two separate tables?

In case this is useful, here's some related notes:

- Computed comes from the `runtime` property. When true it means the var is computed at runtime.
- `type` can either be `FILE_MOUNT` or `ENVIRONMENT_VARIABLE`.

Thanks again, and apologies for such a long response :sweat_smile:

jpJuni0r:
Hey @adamconnelly,

thank you very much for the detailled feedback. I've now updated the PR according to your description and I hope, it's now to your satisfaction.

What I didn't implement was masking the secrets in the table output — I'm not sure, whether you have a convention here. Another thing I noticed is, that we denormalized the data by flattening the hierachy of the context and the configElement. Therefore, the output of the CLI doesn't strictly conform to the model used in the GraphQL API.

internal/cmd/stack/environment.go
```diff
@@ -30,6 +31,51 @@ type ConfigInput struct {
...
+		ContextName string `graphql:"contextName" json:"contextName,omitempty"`
+	} `graphql:"context" json:"context"`
+	Element configElement `graphql:"element" json:"element,omitempty"`
+}
+
+type listEnvQuery struct {
+	Stack struct {
+		// The attachedContext field may not be needed here. It's just kept for reference and to keep the
+		// implementation consistent with the frontend.
+		AttachedContexts []attachedContexts `graphql:"attachedContexts" json:"attachedContexts"`
```
adamconnelly, line 72: Personally I'd just remove this if it isn't used. I just left it in the example query in case I was wrong about it not being required. Apologies if this caused confusion.

internal/cmd/stack/environment.go
```diff
@@ -73,6 +119,89 @@ func setVar(cliCtx *cli.Context) error {
...
+	var query listEnvQuery
+	variables := map[string]interface{}{
+		"stack": graphql.ID(stackID),
+	}
+
+	if err := authenticated.Client.Query(cliCtx.Context, &query, variables); err != nil {
+		return err
+	}
+
+	var outputs []listEnvElementOutput
```
adamconnelly, line 139: nit: can you rename this to `elements` or something like that?

internal/cmd/stack/environment.go
```diff
@@ -73,6 +119,89 @@ func setVar(cliCtx *cli.Context) error {
...
+	for _, attachedContext := range query.Stack.AttachedContexts {
```
adamconnelly, line 140: I think we should remove this - we end up duplicating the variables because they're contained in both `query.Stack.AttachedContexts` and `query.Stack.RuntimeConfig`. For example this is what I see when I run it:

image

internal/cmd/stack/environment.go
```diff
@@ -73,6 +119,89 @@ func setVar(cliCtx *cli.Context) error {
...
+}
+
+func (e *listEnvCommand) showOutputsTable(outputs []listEnvElementOutput) error {
+	tableData := [][]string{{"Name", "Type", "Value", "File Mode"}}
+	for _, output := range outputs {
+		var row []string
+
+		row = append(row, output.Name)
+		row = append(row, output.Type)
+		if output.Value != nil {
```
adamconnelly, line 170: I think we might need special handling for mounted files here. In the case of mounted files, the value is the base64-encoded content of the file. Depending on the file this could be huge.

I've got a couple of ideas:

1. (the simple option) just don't show the value for files in the table output (but still include it in the json output).
2. (the more complex option) decode the value if present, and only show the first X characters.

Personally I'd just go for option 1 and not bother showing the file content in the table output format.

internal/cmd/stack/environment.go
```diff
@@ -73,6 +119,89 @@ func setVar(cliCtx *cli.Context) error {
...
+		return e.showOutputsTable(outputs)
+	case cmd.OutputFormatJSON:
+		return e.showOutputsJSON(outputs)
+	}
+
+	return fmt.Errorf("unknown output format: %v", outputFormat)
+}
+
+func (e *listEnvCommand) showOutputsTable(outputs []listEnvElementOutput) error {
+	tableData := [][]string{{"Name", "Type", "Value", "File Mode"}}
```
adamconnelly, line 164: Can you add a column header for the "Context" column please?

internal/cmd/stack/stack.go
```diff
@@ -151,6 +151,16 @@ func Command() *cli.Command {
 						Before:    authenticated.Ensure,
 						ArgsUsage: "NAME VALUE",
 					},
+					{
+						Name:  "list",
+						Usage: "Lists all the env vars and mounted files for a stack.",
```
adamconnelly, line 156:
```suggestion
						Usage: "Lists all the environment variables and mounted files for a stack.",
```

adamconnelly (commented):
@jpJuni0r thanks a lot for making those changes. I've left some more feedback, but in general it's looking great.

RE denormalization: I think it's fine to denormalise the output like you've done.

RE secret values and masking: I've added a comment about the output for mounted files, but here's a more generic approach:

1. If the `Runtime` property is true, output `<computed>`.
2. If `WriteOnly` is true, output `*****`.
3. If the type is a file, just don't output a value.

Thanks again!

internal/cmd/stack/environment.go
jpJuni0r: Okay, I've removed this part.

internal/cmd/stack/environment.go
jpJuni0r: Sure

internal/cmd/stack/environment.go
jpJuni0r: Okay, I've removed this part as mentioned in a comment above.

internal/cmd/stack/environment.go
jpJuni0r: Okay. I'm now decoding files and will apply special formatting formatting for them. I'm now replacing newlines with spaces and I'm trimming the length. In a future release you may want to trim all long columns in a table.

internal/cmd/stack/environment.go
jpJuni0r: Thanks, that's now fixed.

internal/cmd/stack/stack.go
jpJuni0r: Done.

internal/cmd/stack/environment.go
```diff
@@ -73,6 +117,132 @@ func setVar(cliCtx *cli.Context) error {
...
+
+		row = append(row, output.Name)
+		row = append(row, output.Type)
+
+		var value string
+		switch {
+		case output.Runtime:
+			value = "*****"
+		case output.WriteOnly:
+			value = "<computed>"
```
adamconnelly, line 173: These two are flipped - `Runtime` means the value is calculated at runtime (i.e. `<computed>`), and `WriteOnly` means it's a secret.

internal/cmd/stack/environment.go
```diff
@@ -30,6 +32,48 @@ type ConfigInput struct {
...
+		ID          string `graphql:"id" json:"id,omitempty"`
+		ContextName string `graphql:"contextName" json:"contextName,omitempty"`
+	} `graphql:"context" json:"context"`
+	Element configElement `graphql:"element" json:"element,omitempty"`
+}
+
+type listEnvQuery struct {
+	Stack struct {
+		// The attachedContext field may not be needed here. It's just kept for reference and to keep the
+		// implementation consistent with the frontend.
```
adamconnelly, line 70: I think this comment can be removed now.

internal/cmd/stack/environment.go
adamconnelly: Ok - sounds good.

internal/cmd/stack/environment.go
jpJuni0r: I see, the values are now switched.

internal/cmd/stack/environment.go
jpJuni0r: Sure 👍

internal/cmd/stack/environment.go
```diff
@@ -30,6 +32,46 @@ type ConfigInput struct {
...
+	Name     string  `json:"name"`
+	Type     string  `json:"type"`
+	Value    *string `json:"value"`
+	FileMode *string `json:"fileMode"`
+	// Context specifies the name of the context.
+	Context *string `json:"context"`
+	// Runtime is not printed, it's just used to determine output formatting.
+	Runtime bool
+	// WriteOnly is not printed, it's just used to determine output formatting.
+	WriteOnly bool
```
adamconnelly, line 56: Can you add json struct tags to `Runtime` and `WriteOnly`. They aren't displayed in the output, but they're used in the JSON output. Otherwise we end up with a mixture of cases.

internal/cmd/stack/environment.go
```diff
@@ -73,6 +115,132 @@ func setVar(cliCtx *cli.Context) error {
...
+		}
+	}
+
+	return listEnvElementOutput{
+		Name:     e.ID,
+		Type:     string(e.Type),
+		Value:    value,
+		FileMode: e.FileMode,
+		Context:  contextName,
+		Runtime:  e.Runtime,
```
adamconnelly, line 220: You're not setting `WriteOnly` here, so secrets aren't shown as masked properly in the table output.

internal/cmd/stack/environment.go
```diff
@@ -73,6 +115,132 @@ func setVar(cliCtx *cli.Context) error {
...
+}
+
+func (o *listEnvElementOutput) trimmedValue() string {
+	if o.Value == nil {
+		return ""
+	} else {
+		var result = *o.Value
+
+		valueNoCr := strings.ReplaceAll(*o.Value, "\r", "")
+		valueNoNewlines := strings.ReplaceAll(valueNoCr, "\n", " ")
```
adamconnelly, line 231: If the value isn't longer than the max length, the newlines aren't stripped correctly. Here's a test case highlighting it:

```go
{input: "abc\rdef\nghi\r\njkl", expected: "abc def ghi jkl"},
```

adamconnelly (commented):
@jpJuni0r I've just done some testing and noticed a few issues. I've added a few more comments.

internal/cmd/stack/environment.go
jpJuni0r: Sure, I've updated the code.

internal/cmd/stack/environment.go
jpJuni0r: Good catch, it's now fixed.

internal/cmd/stack/environment.go
jpJuni0r: Good catch, I've added your test case and adjusted the code.

adamconnelly (approved):

PR #73, opened 2022-08-23
Remove FileMode from environment list

adamconnelly:
We've decided to remove the file mode functionality for now, so I'm removing it from spacectl before it's removed from the API.


peterdeme (approved):
lgtm

PR #76, opened 2022-08-24
feat(spacectl/stack): add stack lock and unlock commands

jpJuni0r:


internal/cmd/stack/stack.go
```diff
@@ -223,6 +223,27 @@ func Command() *cli.Command {
...
+			{
+				Name:  "lock",
+				Usage: "Locks a stack for exclusive use.",
+				Flags: []cli.Flag{
+					flagStackID,
+					flagStackLockNote,
+				},
+				Action:    lock,
+				Before:    authenticated.Ensure,
+				ArgsUsage: "NAME",
```
adamconnelly, line 235: I think you actually want to use `ArgsUsage: cmd.EmptyArgsUsage` here. Same with the unlock command (on the grounds they both only use flags).

adamconnelly (commented):
Looks great - just spotted a tiny issue with the arg usage.

internal/cmd/stack/stack.go
jpJuni0r: Of course, it's now fixed.

PR #82, opened 2022-08-26
Add approval and rejection support

adamconnelly:
- Added the ability to approve and reject runs and tasks.
- The approval and rejection mutations accept an optional run ID (if not specified they target the stack blocker). Because of that I've created an extra optional run ID flag and renamed the required one.


internal/cmd/stack/run_review.go
```diff
@@ -0,0 +1,64 @@
+package stack
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/shurcooL/graphql"
```
michieldewilde, line 7: do we do import ordering here?

internal/cmd/stack/stack.go
```diff
@@ -35,13 +35,39 @@ func Command() *cli.Command {
...
 					flagTail,
 				},
 				Action:    runDiscard(),
 				Before:    authenticated.Ensure,
 				ArgsUsage: cmd.EmptyArgsUsage,
 			},
+			{
+				Category: "Run management",
+				Name:     "approve",
+				Usage:    "Approves a run or task. If no run is specified, the approval will be added to the current stack blocker.",
```
michieldewilde, line 48: `Approves a run or task. If no run is specified, the approval will be added to the current stack blocker.` -> `Approves a run or task. If no run is specified, the current stack blocker will be approved.`

michieldewilde (approved):

internal/cmd/stack/run_review.go
adamconnelly: I'll fix this for now, but I think we need to setup golangci-lint on this repo like we do elsewhere.

PR #87, opened 2022-09-02
Use Go 1.19 and bump GitHub Actions versions

peterdeme:


.github/workflows/release.yml
```diff
@@ -9,27 +9,27 @@ jobs:
...
       - name: Set up Go
-        uses: actions/setup-go@v2
-        with: { go-version: 1.18 }
+        uses: actions/setup-go@v3
+        with: { go-version: 1.19 }
 
       - name: Import GPG key
         id: import_gpg
-        uses: paultyng/ghaction-import-gpg@v2.1.0
+        uses: crazy-max/ghaction-import-gpg@v5
```
peterdeme, line 21: The paultyng action is deprecated and archived

.github/workflows/security.yml
```diff
@@ -35,14 +35,14 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout
 
       - name: Run GoSec Security Scanner
-        uses: securego/gosec@master
+        uses: securego/gosec
```
peterdeme, line 41: my rationale of removing `@master` tags from everywhere is that we can't make sure they keep using `master` branch instead of `main`. My assumption is that not using anything will make it so that it uses the default branch whatever it is.

.github/workflows/lint.yml
```diff
@@ -8,11 +8,11 @@ jobs:
   lint:
     name: Lint the code
     runs-on: ubuntu-latest
-    container: golang:1.18
+    container: golang:1.19
 
     steps:
       - name: Check out repository code
-        uses: actions/checkout@master
```
peterdeme, line 15: btw, this repo uses `main` anyway

adamconnelly (approved):

PR #94, opened 2022-09-17
Keep env variable consistency

tomasmik:
Terraform provider uses `SPACELIFT_API_KEY_ENDPOINT` while this used `SPACELIFT_API_ENDPOINT` to set the same value in the env. Now both tools use `SPACELIFT_API_KEY_ENDPOINT`.

For now we still fallback to `SPACELIFT_API_ENDPOINT` if the new value is not set.

### DIFF
--- a/client/session/from_environment.go
+++ b/client/session/from_environment.go
@@ -10,7 +10,13 @@ import (
 const (
 	// EnvSpaceliftAPIEndpoint represents the name of the environment variable
 	// pointing to the Spacelift API endpoint.
+	//
+	// Deprecated
 	EnvSpaceliftAPIEndpoint = "SPACELIFT_API_ENDPOINT"
+
+	// EnvSpaceliftAPIKeyEndpoint represents the name of the environment variable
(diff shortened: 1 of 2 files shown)


marcinwyszynski:
Task linked: [CU-2fdtpn5 Inconsistent use of `SPACELIFT_API_KEY_ENDPOINT` variable between tf provider & spacectl](https://app.clickup.com/t/2fdtpn5)

client/session/from_environment.go
```diff
@@ -40,9 +46,13 @@ func FromEnvironment(ctx context.Context, client *http.Client) func(func(string)
 			return FromAPIToken(ctx, client)(token)
 		}
 
-		endpoint, ok := lookup(EnvSpaceliftAPIEndpoint)
```
peterdeme, line 43: `if ok` then `show warning to not use it anymore` maybe?

b4k3r (approved):
plus Peter's comment

peterdeme (approved):

adamconnelly (approved):

PR #96, opened 2022-09-19
Add trailing newline for set-current-commit output

007:
When updating multiple stacks, the current print format ends up as:
```
Current commit set to "$MESSAGE": (SHA $HASH)Current commit set to "$MESSAGE": (SHA $HASH)Current commit set to "$MESSAGE": (SHA $HASH)
```

Versus after this fix:
```
Current commit set to "$MESSAGE": (SHA $HASH)
Current commit set to "$MESSAGE": (SHA $HASH)
Current commit set to "$MESSAGE": (SHA $HASH)
```

This also removes the sole use of `fmt.Fprintf` in the codebase.

### DIFF
--- a/internal/cmd/stack/set_current_commit.go
+++ b/internal/cmd/stack/set_current_commit.go
@@ -40,7 +40,7 @@ func setCurrentCommit(cliCtx *cli.Context) error {
 		return errors.New("no tracked commit set on the Stack")
 	}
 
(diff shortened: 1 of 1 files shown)


adamconnelly:
@007 thanks for the contribution. The change LGTM, but can you remove the unused import to fix the checks please?

peterdeme (approved):

PR #100, opened 2022-09-30
run_logs improvements

abennett:
There are two parts to this PR:
1. Provide more specific errors for missing struct fields.
2. Lowercase stack names before querying.

Side note, should `run` names also be lowercase?


abennett:
That totally makes sense, @adamconnelly. With this new info, maybe a better change would be to include the stack ID in the `stack list` output to make it more discoverable in the CLI?

adamconnelly:
> maybe a better change would be to include the stack ID in the stack list table output to make it more discoverable in the CLI?

Yes - that makes perfect sense. In fact, we recently had another customer asking for exactly that.

Are you interested in making that change @abennett? If so, feel free to just update this PR since it should be such a small change.

abennett:
Can do!

abennett:
Done!

adamconnelly (approved):

PR #105, opened 2022-11-29
Private Terraform providers - uploading a new version

marcinwyszynski:
For anyone who notices this PR:

- yes, we are working on a private provider registry;
- no, it's not ready;
- if you want to be a beta tester, you know how to find us;
- if you don't know how to find us, no, you cannot be a beta tester;


internal/cmd/provider/create_version.go
```diff
@@ -0,0 +1,144 @@
...
+
+		if err := authenticated.Client.Mutate(cliCtx.Context, &changelogMutation, variables); err != nil {
+			return errors.Wrap(err, "could not update changelog")
+		}
+
+		return nil
+	}
+}
+
+func registerVersion(ctx context.Context, dir string, versionID string, artifact *GoReleaserArtifact) error {
```
adamconnelly, line 126: nit: maybe `registerPlatform` or `registerArtifact`?

internal/cmd/provider/flags.go
```diff
@@ -0,0 +1,28 @@
...
+var flagGoReleaserDir = &cli.StringFlag{
+	Name:  "goreleaser-dir",
+	Usage: "Directory containing the GoReleaser build artifacts",
+	Value: "dist",
+}
+
+var flagSigningKeyID = &cli.StringFlag{
+	Name:     "siging-key-id",
+	Usage:    "ID of the signing key used to sign the provider version",
+	EnvVars:  []string{"GPG_FINGERPRINT"},
```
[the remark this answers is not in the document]
marcinwyszynski: Re: 1
That's because this is the environment variable that GoRelease uses for signing. If you run it in a single CI/CD pipeline (which most folks probably will), the will be no need for setting this explicitly.

Re: 2
I'll rename the flag to match the environment variable (`gpg-fingerprint`)

internal/cmd/provider/create_version.go
marcinwyszynski: Fair point.

marcinwyszynski: Done.

internal/cmd/provider/create_version.go
```diff
@@ -0,0 +1,165 @@
...
+			CreateTerraformProviderVersion struct {
+				SHA256SumsUploadURL    string `graphql:"sha256SumsUploadURL"`
+				SHA256SumsSigUploadURL string `graphql:"sha256SumsSigUploadURL"`
+				Version                struct {
+					ID string `graphql:"id"`
+				} `graphql:"version"`
+			} `graphql:"terraformProviderVersionCreate(provider: $provider, input: $input)"`
+		}
+
+		protocolVersions := []string{"5.0"}
```
TheMacies, line 62: Should this maybe be a default value for the flag? This way, it would be automatically visible in usage print.

internal/cmd/provider/create_version.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/urfave/cli/v2"
+
+	"github.com/spacelift-io/spacectl/internal/cmd/authenticated"
+)
+
+func createVersion() cli.ActionFunc {
+	return func(cliCtx *cli.Context) error {
+		// Assuming that spacectl is ran from the root of the repository,
+		// containing the release artifacts in the "dist" directory.
+		dir := "dist"
```
TheMacies, line 18: Should this maybe be a default value for the flag? This way, it would be automatically visible in usage print.

internal/cmd/provider/goreleaser.go
```diff
@@ -0,0 +1,219 @@
...
+	Changelog *string
+}
+
+// BuildGoReleaserVersionData builds a GoReleaserMetadata struct by looking at the
+// distribution directory created by GoReleaser.
+func BuildGoReleaserVersionData(dir string) (*GoReleaserVersionData, error) {
+	var out GoReleaserVersionData
+
+	// Read the artifacts file.
+	artifactsPath := dir + "/artifacts.json"
```
TheMacies, line 33: It's safer to use filepath.Join as dir might be a custom value provided by the user

internal/cmd/provider/goreleaser.go
```diff
@@ -0,0 +1,219 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to read %s", artifactsPath)
+	}
+
+	if err := json.Unmarshal(artifactsData, &out.Artifacts); err != nil {
+		return nil, errors.Wrapf(err, "failed to unmarshal %s", artifactsPath)
+	}
+
+	// Read the metadata file.
+	metadataPath := dir + "/metadata.json"
```
TheMacies, line 46: It's safer to use filepath.Join as dir might be a custom value provided by the user

internal/cmd/provider/goreleaser.go
```diff
@@ -0,0 +1,219 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to read %s", metadataPath)
+	}
+
+	if err := json.Unmarshal(metadataData, &out.Metadata); err != nil {
+		return nil, errors.Wrapf(err, "failed to unmarshal %s", metadataPath)
+	}
+
+	// Read the CHANGELOG.
+	changelogPath := dir + "/CHANGELOG.md"
```
TheMacies, line 59: It's safer to use filepath.Join as dir might be a custom value provided by the user

TheMacies (approved):
Just 2 nits. I copied them over to similar occurrences. Other than that LGTM

internal/cmd/provider/create_version.go
marcinwyszynski: Does this library even support default values?

TheMacies: Oh, true. The v2.16.3 that we're using seems not to support them. I was looking at the cli repo, and it seems that v2.23.5 already does.

marcinwyszynski: OK I can try to bump the library, using default values will obviously be much cleaner.

internal/cmd/provider/create_version.go
marcinwyszynski: Done

internal/cmd/provider/goreleaser.go
marcinwyszynski: Done

internal/cmd/provider/goreleaser.go
marcinwyszynski: Done.

PR #107, opened 2022-12-17
stack: add enable + disable commands

lexton:


internal/cmd/stack/enable.go
```diff
@@ -0,0 +1,52 @@
...
+		return fmt.Errorf("expected zero arguments but got %d", nArgs)
+	}
+
+	var mutation stackDisableMutation
+	variables := map[string]interface{}{
+		"stack": graphql.ID(stackID),
+	}
+
+	return authenticated.Client.Mutate(cliCtx.Context, &mutation, variables)
+}
```
peterdeme, line 52: How about this? A bit less code duplication.

```go
func enable(cliCtx *cli.Context) error {
	return enableDisable[stackEnableMutation](cliCtx)
}

func disable(cliCtx *cli.Context) error {
	return enableDisable[stackDisableMutation](cliCtx)
}

func enableDisable[T any](cliCtx *cli.Context) error {
	stackID := cliCtx.String(flagStackID.Name)

	if nArgs := cliCtx.NArg(); nArgs != 0 {
		return fmt.Errorf("expected zero arguments but got %d", nArgs)
	}

	var mutation T
	variables := map[string]interface{}{
		"stack": graphql.ID(stackID),
	}

	return authenticated.Client.Mutate(cliCtx.Context, &mutation, variables)
}
```

peterdeme:
@lexton thanks for the contribution 🙏

internal/cmd/stack/enable.go
lexton: Sounds good - I'm not fussed, easy enough to keep clean. I copied the lock.go code to implement this.

peterdeme:
Released in [v0.16.0](https://github.com/spacelift-io/spacectl/releases/tag/v0.16.0)

PR #116, opened 2023-02-02
Try to listen on random port for login command

tomasmik:
We will still fallback to the old port if this doesn't work for some reason.
It's better to do this rather than pick a random port, at least then it's easily addressable by the user.

**Requires a backend and frontend update.**


marcinwyszynski:
Task linked: [CU-862j2cj1e spacectl web Login broken if another service is bound on port 8020](https://app.clickup.com/t/862j2cj1e)

Yantrio (approved):

Yantrio:
Note: finding open ports then binding is always susceptible to race conditions, but this should be fine 👍

c032:
This PR is already approved, but I want to note that a possible alternative is to make the `getOpenPort` function return a `net.Listener`, and use _that_ with `server.Serve(l)` (instead of `server.ListenAndServe()`).

(The `getOpenPort` function would need a different name in this case, though.)

That way we never actually close the listener we already have.

For reference, `(*http.Server).ListenAndServe` [doesn't do much more than that](https://github.com/golang/go/blob/go1.20/src/net/http/server.go#L2976-L2989).

internal/cmd/profile/login_command.go
```diff
@@ -303,3 +302,32 @@ func persistAccessCredentials(creds *session.StoredCredentials) error {
 		Credentials: creds,
 	})
 }
+
+func serveOnOpenPort(handler func(w http.ResponseWriter, r *http.Request)) (*http.Server, int, error) {
+	addr, err := net.ResolveTCPAddr("tcp", ":0")
```
tomasmik, line 307: Tested this on windows to see if it actually works: 
image

Yantrio (approved):

Issue #54, opened 2022-04-01
Implement spacectl open


marcinwyszynski:
Feature request: `spacectl open` -- when run inside a git repo, open the corresponding Spacelift stack in my browser.
(Bonus #1: if there are multiple matching stacks, show a menu/selector)
(Bonus #2: remember my choice based on git repo and subdir within (for monorepos), so I only have to select once)

tomasmik:
Done in: https://github.com/spacelift-io/spacectl/pull/117

PR #117, opened 2023-02-10
Add a command to open a stack in your browser

tomasmik:
How this works:

You can navigate to any repository on your local machine and execute `stack open` which will search for a stack which was created from a git repository you're currently in and has a project root set to your current subdir.

If more than 5 results are returned a very basic search function is invoked and you are able to either filter the returned results or still use arrow keys to select.

The returned list is never longer than 10 stacks and scrolling is automatically enabled if the list is longer than that. If only 1 stack is found, we don't display anything and instantly open a browser window.

Using params you can:
* Change the count of stacks returned (only 10 are shown anyway before scrolling is enforced)
* Ignore current subdir and return all stacks for the repository
* Take current branch in to account when searching for stacks

**Caveats**

* If users have a lot of stacks from the same repo+subdir they will have to adjust count parameter and filter. Though I suspect this won't be typical.
* `git` must be installed on the machine for this to work
* Not sure how windows defender would treat these `.Exec` commands. This should be fine, but you never know.

https://github.com/spacelift-io/spacectl/issues/54

### DIFF
--- a/internal/browser.go
+++ b/internal/browser.go
@@ -1,0 +1,38 @@
+package internal
+
+import (
+	"os/exec"
+	"runtime"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// OpenWebBrowser will try to open the default browser
+// on a the callers machine. It supports windows, darwin and windows.
+func OpenWebBrowser(url string) error {
+	var cmd *exec.Cmd
+	switch runtime.GOOS {
+	case "linux":
+		cmd = exec.Command("xdg-open", url)
+	case "darwin":
+		cmd = exec.Command("open", url)
+	case "windows":
+		r := strings.NewReplacer("&", "^&")
+		cmd = exec.Command("cmd", "/c", "start", r.Replace(url)) //#nosec
+	default:
+		return errors.New("unsupported platform")
+	}
+
+	err := cmd.Start()
+	if err != nil {
+		return errors.Wrap(err, "could not open the browser")
+	}
+
+	err = cmd.Wait()
+	if err != nil {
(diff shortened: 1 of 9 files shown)


Yantrio (approved):

internal/browser.go
```diff
@@ -0,0 +1,38 @@
...
+	"os/exec"
+	"runtime"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// OpenWebBrowser will try to open the default browser
+// on a the callers machine. It supports windows, darwin and windows.
+func OpenWebBrowser(url string) error {
```
cube2222, line 13: There's actually a nice little package we could use: https://github.com/pkg/browser

internal/cmd/stack/open_command.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	name, err := getRepositoryName()
+	if err != nil {
+		return err
+	}
+
+	return findAndSelect(cliCtx.Context, &stackSearchParams{
+		count:          count,
+		projectRoot:    subdir,
```
cube2222, line 51: Does this correctly interact with empty project roots?

cube2222 (approved):
with nits

internal/browser.go
tomasmik: I would rather open a ticket and implement this in a separate ticket as it affects other parts of spacectl too 🤷🏻

internal/cmd/stack/open_command.go
tomasmik: I have stacks: 
image

If I execute open in dir of a repo I get:
image

PR #121, opened 2023-03-01
Allow to bind a specific port for login command

tomasmik:
By default we bind to a random port, but we still allow to specify a port if needed. I feel like this makes more sense than going the other way around.


Clickup-user:
Task linked: [CU-861mea8zh Allow to bind spacectl browser to a specific port](https://app.clickup.com/t/861mea8zh)

cube2222 (approved):
LGTM, let's get this released today!

PR #122, opened 2023-03-01
Provide a flag to print additional stack data for table format

tomasmik:


Clickup-user:
Task linked: [CU-2vdb0bd Beamer FR: Spacelift CLI - show stack labels on output](https://app.clickup.com/t/2vdb0bd)

internal/cmd/stack/flags.go
```diff
@@ -99,3 +99,10 @@ var flagSearchCount = &cli.IntFlag{
 	Usage: "[Optional] Indicate the maximum count of elements returned when searching stacks",
 	Value: 30,
 }
+
+var flagVerbose = &cli.BoolFlag{
+	Name:     "verbose",
```
adamconnelly, line 104: I wonder if we'd be better just calling it `show-labels` for now given that's the only additional field we're adding? That ways it's really clear what it does, and doesn't get mistaken for something that sets the log level or something like that (on the grounds quite a few commands have a `verbose` option to increase the logging).

I suspect we're unlikely to want to include tons of extra columns in the table output because it very quickly becomes unreadable.

internal/cmd/stack/flags.go
```diff
@@ -99,3 +99,10 @@ var flagSearchCount = &cli.IntFlag{
 	Usage: "[Optional] Indicate the maximum count of elements returned when searching stacks",
 	Value: 30,
 }
+
+var flagShowLabels = &cli.BoolFlag{
+	Name:     "show-labels",
+	Usage:    "[Optional] Indicate whetever an additional labels column should be printed when outputting stack data in a table format",
```
adamconnelly, line 105:
```suggestion
	Usage:    "[Optional] Indicates that stack labels should be printed when outputting stack data in the table format",
```

adamconnelly (approved):
Just need to tweak the help text slightly. But other than that LGTM.

PR #109, opened 2022-12-20
login: web-based login should bind to localhost

lexton:
This shortcuts the macos warning and confirmation to open network connections.

The origin of these requests should only ever come from localhost.

PR #126, opened 2023-03-21
login: bind to localhost for web based login's

lexton:
regression fix for: https://github.com/spacelift-io/spacectl/pull/109


tomasmik:
Looks fine and while it should work it would be nice if someone with linux could test this. @adamconnelly could you give us a hand here?

adamconnelly:
> Looks fine and while it should work it would be nice if someone with linux could test this. @adamconnelly could you give us a hand here?

@tomasmik LGTM - tested it on Ubuntu and it works fine for me.

tomasmik (approved):

Issue #60, opened 2022-05-06
Feature Request: default stack id per directory


zxaos:
I'd be really nice not to have to specify the stack id for many commands and instead have it read from or known location in a file or even an environment variable!

tomasmik:
This kinda exist right now. Currently if no `--id` is provided, spacectl will attempt to detect a stack based on the current directory (if it's a git repository)

PR #128, opened 2023-03-24
Test for empty subdir when searching for stacks

bodgit:
I was looking to see if it was possible to default the stack ID to an environment variable and found #60 but also found #125 had recently been merged so it should figure it out from the underlying cloned git repository. A new release hadn't been cut with this functionality so I tried building from source. I found `spacectl` still couldn't figure out the stack ID automatically.

All of my stacks map 1:1 with their own separate repository so the project root is unset. I found I needed to apply this small patch in order for the search to return any results.

Also, it currently always prompts me to pick the stack from the search results, even if there's only ever 1 stack returned. This looks to be caused by this condition:

https://github.com/spacelift-io/spacectl/blob/d36ea753da14b0bbfcc6f5c3ff2003fbb2800ce6/internal/cmd/stack/stack_selector.go#L68

`forcePrompt` is always `true` so it doesn't matter how many search results have been returned.

### DIFF
--- a/internal/cmd/stack/open_command.go
+++ b/internal/cmd/stack/open_command.go
@@ -148,7 +148,7 @@ func searchStacks(ctx context.Context, p *stackSearchParams) ([]stack, error) {
 		},
 	}
 
-	if p.projectRoot != nil {
+	if p.projectRoot != nil && *p.projectRoot != "" {
 		root := strings.TrimSuffix(*p.projectRoot, "/")
 		conditions = append(conditions, structs.QueryPredicate{
 			Field: graphql.String("projectRoot"),


tomasmik:
Sorry for the late reply.


> Also, it currently always prompts me to pick the stack from the search results, even if there's only ever 1 stack returned. This looks to be caused by this condition

that is done on purpose, so you don't accidentally execute an action on a wrong stack. This PR however is alright and if you would solve the conflicts, I'm happy to approve it.

bodgit:
> that is done on purpose, so you don't accidentally execute an action on a wrong stack.

It would be nice if there was a way to turn that off, either a flag or by setting an environment variable maybe?

> This PR however is alright and if you would solve the conflicts, I'm happy to approve it.

Conflicts fixed.

tomasmik:
> It would be nice if there was a way to turn that off, either a flag or by setting an environment variable maybe?

Feel free to open an issue for that.

tomasmik (approved):

PR #133, opened 2023-03-28
local-preview: allow packing only of the project-root

lexton:
This is important to us because we use a mono repo, and we don't want to keep uploading the whole repo for each local run. This takes on average an extra 10s per run, and is problematic in low bandwidth environments.

This PR does a few things:

* adds a trailing `/` to the project root. I'm not sure what the expected behaviour is here. We could also remove the trailing slashes from the project root configurations of our stacks - but would be great if it was normalized so that this wasn't an issue.
* Allows a `project-root-only` flag that will only pack the project root for uploads. This is a 2000x reduction in upload size for our monorepo.
* fixes a bug in the `.git` detection algo where it has the side effect of also adjusting the project root.

### DIFF
--- a/internal/cmd/module/local_preview.go
+++ b/internal/cmd/module/local_preview.go
@@ -51,7 +51,7 @@ func localPreview() cli.ActionFunc {
 
 		fp := filepath.Join(os.TempDir(), "spacectl", "local-workspace", fmt.Sprintf("%s.tar.gz", uploadMutation.UploadLocalWorkspace.ID))
 
-		matchFn, err := internal.GetIgnoreMatcherFn(ctx)
+		matchFn, err := internal.GetIgnoreMatcherFn(ctx, "")
 		if err != nil {
 			return fmt.Errorf("couldn't analyze .gitignore and .terraformignore files")
 		}
--- a/internal/cmd/stack/open_command.go
+++ b/internal/cmd/stack/open_command.go
@@ -100,7 +100,7 @@ func getGitCurrentBranch() (string, error) {
 	return result, nil
 }
 
-// getGitRepositorySubdir will travese the path back to .git
+// getGitRepositorySubdir will traverse the path back to .git
 // and return the path it took to get there.
 func getGitRepositorySubdir() (string, error) {
 	current, err := os.Getwd()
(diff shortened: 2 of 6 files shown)


internal/cmd/module/local_preview.go
```diff
@@ -51,7 +51,7 @@ func localPreview() cli.ActionFunc {
 
 		fp := filepath.Join(os.TempDir(), "spacectl", "local-workspace", fmt.Sprintf("%s.tar.gz", uploadMutation.UploadLocalWorkspace.ID))
 
-		matchFn, err := internal.GetIgnoreMatcherFn(ctx)
+		matchFn, err := internal.GetIgnoreMatcherFn(ctx, "")
```
lexton, line 54: I think we should switch off this behaviour for modules. I know they are generally handled slightly differently be the run side orchestrator?

internal/local_preview.go
```diff
@@ -45,8 +46,10 @@ func MoveToRepositoryRoot() error {
 	}
 }
 
-// GetIgnoreMatcherFn creates an ignore-matcher for archiving purposes. Respects gitignore and terraformignore.
-func GetIgnoreMatcherFn(ctx context.Context) (func(filePath string) bool, error) {
+// GetIgnoreMatcherFn creates an ignore-matcher for archiving purposes
+// This function respects gitignore and terraformignore, and
+// optionally if a projectRoot is provided it only include files from this root
+func GetIgnoreMatcherFn(ctx context.Context, projectRoot string) (func(filePath string) bool, error) {
```
lexton, line 52: There is the open question I have after looking at this code - do we want to be doing recursive lookups of the .gitignore files or just look at the files in the root of the repo? Seems like this would ignore the local directory and go straight to the root.

internal/cmd/stack/open_command.go
```diff
@@ -110,22 +110,16 @@ func getGitRepositorySubdir() (string, error) {
...
-		if err != nil {
-			return "", fmt.Errorf("couldn't get current working directory: %w", err)
-		}
-
-		parent := filepath.Dir(cwd)
-		if err := os.Chdir(parent); err != nil {
-			return "", fmt.Errorf("couldn't set current working directory: %w", err)
-		}
-		root = parent
+	if !strings.HasSuffix(current, "/") {
```
lexton, line 121: I don't know what the correct behavior is here.

Can we assume that the project root always has a trailing slash?

I don't think I have seen any guidance on this in the docs.

internal/cmd/stack/open_command.go
```diff
@@ -110,22 +110,16 @@ func getGitRepositorySubdir() (string, error) {
...
+		root = filepath.Dir(root)
+	}
 
-		cwd, err := os.Getwd()
-		if err != nil {
-			return "", fmt.Errorf("couldn't get current working directory: %w", err)
-		}
-
-		parent := filepath.Dir(cwd)
-		if err := os.Chdir(parent); err != nil {
```
lexton, line 125: This to me - looks like a bug - I don't think looking up the sub dir should change the current working directory - my implementation is a fix for this - but there still may be a better way of detecting this. This also may have some weird side effect if it is loaded as a git submodule.

lexton (commented):
Just my initial thoughts on my own code.

PR #134, opened 2023-03-29
getGitRepositorySubdir: shouldn't have the side effect of changing dir

lexton:
This keeps the directory the same and will let us look up the subdir multiple times.

`getStackID` has the unfortunate side effect of changing the current working directory of the current process - which means we can't actually use the lookup function again later potentially for looking up the current project root.


cube2222 (dismissed):
LGTM, thanks for the contribution!

cube2222 (approved):

adamconnelly (approved):
Makes sense to me.

PR #135, opened 2023-03-29
searchStacks: append a trailing '/' to search func

lexton:
## Problem

If there is currently a trailing `/` in the lookup path - it will not find the project root.

```
$ spacectl stack list -o json | jq -r '.[] | [.id, .projectRoot] | @csv' | head -n1
"spacelift-demo","spacelift/demo/"
```

Example failing lookup:
```
$ cd spacelift/demo
$ spacectl stack local-preview 
2023/04/04 09:40:47 no --id flag was provided and stack could not be found by searching the current directory
```

On further debugging:

When I am in the `spacelift/demo` directory, the call to `getGitRepositorySubdir` will always return `spacelift/demo`

However `findAndSelectStack` does will not match `spacelift/demo` to the stack on the server side which is stored as `spacelift/demo/`.

## Solution

This solution will also run a search for `[]string{"spacelift/demo", "spacelift/demo/"}` and do an exact match - so it still won't be fuzzy, but would still allow us to find the relevant stacks.

### DIFF
--- a/internal/cmd/stack/open_command.go
+++ b/internal/cmd/stack/open_command.go
@@ -149,10 +149,11 @@ func searchStacks(ctx context.Context, p *stackSearchParams) ([]stack, error) {
 	}
 
 	if p.projectRoot != nil {
+		root := strings.TrimSuffix(*p.projectRoot, "/")
 		conditions = append(conditions, structs.QueryPredicate{
 			Field: graphql.String("projectRoot"),
 			Constraint: structs.QueryFieldConstraint{
-				StringMatches: &[]graphql.String{graphql.String(*p.projectRoot)},
+				StringMatches: &[]graphql.String{graphql.String(root), graphql.String(root + "/")},
 			},
 		})
 	}


internal/cmd/stack/open_command.go
```diff
@@ -149,10 +149,17 @@ func searchStacks(ctx context.Context, p *stackSearchParams) ([]stack, error) {
 	}
 
 	if p.projectRoot != nil {
+		roots := []graphql.String{graphql.String(*p.projectRoot)}
```
cube2222, line 152: @lexton Just to make sure this is resistant in face of changes in other parts of the repo, could you strip the "/" with strings.TrimSuffix on the project root here?

I know it doesn't have the trailing slash right now and will do nothing, but this wouldn't be fun to debug if it changed in the future.

lexton: Yeah - totally agree - that's totally the right call :D

internal/cmd/stack/open_command.go
```diff
@@ -149,10 +149,16 @@ func searchStacks(ctx context.Context, p *stackSearchParams) ([]stack, error) {
 	}
 
 	if p.projectRoot != nil {
+		// if we are at the root (e.g. "") or if we have a provided path we want to search both with and without trailing slash
+		roots := []graphql.String{graphql.String(strings.TrimSuffix(*p.projectRoot, "/"))}
+		if len(roots[0]) == 0 || roots[0][len(roots[0])-1] != '/' {
+			roots = append(roots, graphql.String("/"))
```
cube2222, line 155: I think this should be
```
roots = append(roots, roots[0] + graphql.String("/"))
```

cube2222 (approved):

PR #136, opened 2023-04-04
lightweight packaging: only upload the project_root

lexton:
## Lightweight Packaging

Use cases, where we know we only need to stack `project_root` to deploy the stack successfully, this will make the upload much more efficient. In our test cases, this takes the upload from 2MB to 1/2kb for most of our stacks, so a 1000x reduction in upload bandwidth.

This is very useful in low bandwidth environments, especially in working with a mono repo.

## Solution

This currently only works for stacks, but it could probably also work with the modules pretty trivially, maybe we would want this to be the default behavior in the modules.

## Testing

I haven't got a stack that uses the root of the repo - so haven't been able to test where a stack has no project root set.

### DIFF
--- a/internal/cmd/stack/flags.go
+++ b/internal/cmd/stack/flags.go
@@ -32,6 +32,12 @@ var flagNoFindRepositoryRoot = &cli.BoolFlag{
 	Value: false,
 }
 
+var flagProjectRootOnly = &cli.BoolFlag{
+	Name:  "project-root-only",
+	Usage: "Indicate if spacelift should only package files inside this stacks project_root",
+	Value: false,
+}
+
 var flagRequiredCommitSHA = &cli.StringFlag{
 	Name:     "sha",
 	Usage:    "[Required] `SHA` of the commit to set as canonical for the stack",
--- a/internal/cmd/stack/local_preview.go
+++ b/internal/cmd/stack/local_preview.go
@@ -22,13 +22,25 @@ func localPreview() cli.ActionFunc {
 		}
 		ctx := context.Background()
 
+		var packagePath *string = nil
+		if cliCtx.Bool(flagProjectRootOnly.Name) {
+			root, err := getGitRepositorySubdir()
+			if err != nil {
+				return fmt.Errorf("couldn't get the packagePath: %w", err)
+			}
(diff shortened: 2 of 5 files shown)


internal/cmd/stack/flags.go
```diff
@@ -32,6 +32,12 @@ var flagNoFindRepositoryRoot = &cli.BoolFlag{
 	Value: false,
 }
 
+var flagProjectRootOnly = &cli.BoolFlag{
+	Name:  "project-root-only",
```
lexton, line 36: Totally open to flag naming - I have zero opinions here - but this made sense to me, so if there is a broader strategy - or a better name I'm all ears.

internal/cmd/stack/local_preview.go
```diff
@@ -22,13 +22,25 @@ func localPreview() cli.ActionFunc {
...
 			if err := internal.MoveToRepositoryRoot(); err != nil {
 				return fmt.Errorf("couldn't move to repository root: %w", err)
 			}
 		}
 
-		fmt.Println("Packing local workspace...")
+		if packagePath == nil {
+			fmt.Println("Packing local workspace...")
+		} else {
+			fmt.Printf("Packing project_root as local workspace... project_root='%s'\n", *packagePath)
```
cube2222, line 42: Maybe just replace `Packing project_root as` with `Packing '%s' as` here, without the `project_root='%s'`? It looks a bit like debug logs otherwise.

internal/cmd/stack/local_preview.go
```diff
@@ -22,13 +22,25 @@ func localPreview() cli.ActionFunc {
...
 			if err := internal.MoveToRepositoryRoot(); err != nil {
 				return fmt.Errorf("couldn't move to repository root: %w", err)
 			}
 		}
 
-		fmt.Println("Packing local workspace...")
+		if packagePath == nil {
+			fmt.Println("Packing local workspace...")
+		} else {
+			fmt.Printf("Packing '%s' as local workspace\n", *packagePath)
```
cube2222, line 42: A bit inconsistent now, with one being Println and trailing dots, while the other is Printf with an explicit newline and no dots 😅

Issue #138, opened 2023-04-06
local-preview: .gitignore only discovered at the root of the repo


lexton:
Currently the `.gitignore` and the `.terraformignore` files are only read at the root of the repo.

This doesn't align with how `git` describes its working `.gitignore` file resolution.

from `man gitignore`: 
>  Patterns read from a .gitignore file in the same directory as the path, or in any parent directory (up to the top-level of the working tree), with patterns in the higher level files being overridden by those in lower level files down to the directory containing the file. These patterns match relative to the location of the .gitignore file. A project normally includes such .gitignore files in its repository, containing patterns for files generated as part of the project build.

wadells:
We tripped over this too.  We had some secrets on the filestyem that were gitignored by `directory/.gitignore`, but were uploaded to spacelift during a `local-preview`.  This was confounded by spacelift's slack integration sharing local-preview presigned s3 url -- effectively sharing these sensitive files with the entire company.

wadells:
Additionally, spacectl should mind `$GIT_DIR/info/exclude` (checkout specific and repo specific excludes) and `git config core.excludesfile` (user specific global excludes).

https://git-scm.com/docs/gitignore#_description

tomasmik:
hey @wadells 

> This was confounded by spacelift's slack integration sharing local-preview presigned s3 url -- effectively sharing these sensitive files with the entire company.

This was fixed.

wadells:
> hey @wadells
> 
> > This was confounded by spacelift's slack integration sharing local-preview presigned s3 url -- effectively sharing these sensitive files with the entire company.
> 
> This was fixed and should not happen again.

Thank you -- we noticed and appreciate that change to the slack integration.

PR #139, opened 2023-04-09
Use all relevant ignore files in archive matcher

c032:
I tried doing #138 to pass time. Most explanation is in the commit message.

But for a quick skim, start at the diff for `internal/cmd/module/local_preview.go` or `internal/cmd/stack/local_preview.go`. Specifically at the `matchFn` function.

I cannot test this because I don't have an account, and the free account sounds like a 14-day trial.

Also I'm probably missing a lot of context from the project so it's likely I'm overcomplicating this.

I hope this PR is at least a bit helpful, but if not, feel free to reject.

---

Also, I could not build the project at first because the required version of `github.com/marcinwyszynski/graphql` didn't exist (!?). I had to fix it before I could even build the `main` branch.


c032:
Abandoning.

PR #150, opened 2023-06-05
Use ID of stack instead of name

### LINKED ISSUE #137: local-preview: returns a Not Found if the slug doesn't match the name
It doesn't seem like it does the slug resolution based on the name, and will return 404's from the server once trying to use the stack name rather than slug.
```
$ cd spacelift/demo
$  git:(lexton/demo-slug) spacectl stack local-preview
✔ Spacelift: demo
Packing local workspace...
2023/04/06 14:21:11 not found
```

leetrout:
This PR uses the ID of the stack (the slug) instead of the name.

Several commands were bugged by this.

Closes #137

### DIFF
--- a/internal/cmd/stack/stack_selector.go
+++ b/internal/cmd/stack/stack_selector.go
@@ -85,7 +85,7 @@ func findAndSelectStack(ctx context.Context, p *stackSearchParams, forcePrompt b
 			return "", err
 		}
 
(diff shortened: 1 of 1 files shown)


adamconnelly (approved):
LGTM. Thanks a lot for fixing that for us! :bow:

PR #153, opened 2023-06-15
Add support for custom host binding with --bind-host flag and environment variables

mcalhoun:
## What

- Add new `--bind-host` flag to allow the user to bind to a host other than `localhost`
- Add the ability to specify the `--bind` flag via `SPACECTL_BIND_PORT` and the new `--bind-host` via `SPACECTL_BIND_HOST` environment variables

## Why

When running `spacectl` in a docker environment such as [Cloud Posse's Geodesic](https://github.com/cloudposse/geodesic), `localhost` isn't available outside of the container, so we have to bind to `0.0.0.0`. The new functionality gives us the ability to do the following:

```bash
# /etc/profile
export SPACECTL_BIND_HOST=0.0.0.0
export SPACECTL_BIND_PORT=$(command to get the mapped port docker is listening on)
```

Then we can just use the `spacectl profile login` anywhere within the container and it will work as expected.

### DIFF
--- a/internal/cmd/profile/flags.go
+++ b/internal/cmd/profile/flags.go
@@ -6,4 +6,5 @@ var flagBindPort = &cli.IntFlag{
 	Name:     "bind",
 	Usage:    "[Optional] specify the port used for binding the server when logging in through a browser",
 	Required: false,
+	EnvVars:  []string{"SPACECTL_BIND_PORT"},
 }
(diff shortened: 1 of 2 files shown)


Yantrio (approved):
I love the approach of having 0 as default so it can simplify serveOnOpenPort :) 

LGTM 👍

tomasmik (approved):

PR #159, opened 2023-07-06
add check for empty environment variables

jwenz723:
I would like to execute some `spacectl` commands in custom workflow steps (pre-init, pre-plan, etc) within some spacelift stacks. In spacelift stacks the `SPACELIFT_API_TOKEN` environment variable exists by default with a `<computed>` value, but unfortunately this value doesn't have sufficient permissions to work with the way spacectl authenticates. Ideally this problem would be resolved, but I don't understand that problem.

Therefore, I want to use the `SPACELIFT_API_KEY_ID` and `SPACELIFT_API_KEY_SECRET` variables in my stack to authenticate spacectl. There isn't a way to unset the `SPACELIFT_API_TOKEN` computed environment variable that exists by default on spacelift stacks. If the `SPACELIFT_API_TOKEN` environment variable exists with an empty value then it will still take precedence over the `SPACELIFT_API_KEY_ID` and `SPACELIFT_API_KEY_SECRET` variables. 

This PR adds code to check if any of the environment variables which spacectl uses to authenticate are set to an empty string. If a variable is set to an empty string then the variable will be ignored and considered as unset. This should allow me to use the `SPACELIFT_API_KEY_ID` and `SPACELIFT_API_KEY_SECRET` variables in my spacelift stacks.

### DIFF
--- a/client/session/from_environment.go
+++ b/client/session/from_environment.go
@@ -54,7 +54,7 @@ func FromEnvironment(ctx context.Context, client *http.Client) func(func(string)
 		}
 
 		endpoint, ok := lookup(EnvSpaceliftAPIKeyEndpoint)
-		if !ok {
+		if !ok || endpoint == "" {
 			// Keep backwards compatibility with older version of spacectl.
 			endpoint, ok = lookup(EnvSpaceliftAPIEndpoint)
 			if !ok {
--- a/client/session/from_environment.go
+++ b/client/session/from_environment.go
@@ -36,6 +36,12 @@ const (
 	EnvSpaceliftAPIGitHubToken = "SPACELIFT_API_GITHUB_TOKEN" // #nosec G101
 )
 
+var (
+	ErrEnvSpaceliftAPIKeyID       = fmt.Errorf("%s missing from the environment", EnvSpaceliftAPIKeyID)
+	ErrEnvSpaceliftAPIKeySecret   = fmt.Errorf("%s missing from the environment", EnvSpaceliftAPIKeySecret)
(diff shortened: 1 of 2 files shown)


client/session/from_environment.go
```diff
@@ -36,40 +36,46 @@ const (
...
+
 // FromEnvironment creates a Spacelift session from the environment.
 func FromEnvironment(ctx context.Context, client *http.Client) func(func(string) (string, bool)) (Session, error) {
 	return func(lookup func(string) (string, bool)) (Session, error) {
 		if lookup == nil {
 			lookup = os.LookupEnv
 		}
 
-		if token, ok := lookup(EnvSpaceliftAPIToken); ok {
+		if token, ok := lookup(EnvSpaceliftAPIToken); ok && token != "" {
```
marcinwyszynski, line 52: Just want to make sure I understand the mechanics here - is the intended usage something like:

```bash
SPACELIFT_API_TOKEN="" spacectl ...
```

marcinwyszynski:
Can you please fix the linter error? You can make these sentinel errors unexported (I can't see them used outside the package, so it's my preffered option), or add a comment to each.

client/session/from_environment.go
jwenz723: Yeah that is my plan. 

The other option would be to have a variable that allows specifying which credential type to be used regardless of which credential env vars exist, but I think the current approach is easier and more clear.

jwenz723:
> Can you please fix the violation reported by the linter? You can make these sentinel errors unexported (I can't see them used outside the package, so it's my preferred option), or add a comment to each.

Done.

adamconnelly (approved):
@jwenz723 Marcin is on holiday this week, so he's asked me to take over.

Your changes look good - thanks a lot for the contribution. :bow:

PR #162, opened 2023-07-20
Always pass state version when getting run logs

tomasmik:
We've added to repeat the same state for a run which requires you to provide the state version when getting logs. This adds the state version to the request.


mbialon (approved):
LGTM. It would be nice if the PR had some information about why we're doing this.

tomasmik:
> LGTM. It would be nice if the PR had some information about why we're doing this.

Fair. Added.

Issue #158, opened 2023-06-30
Command args for `login` command


lohrm-stabl:
We used `spacectl` in short-lived development containers. This is way we have to execute `spacectl profile login` very often.
It would be great if we could supply the URL and authentication kind via command args. Its quite tedious to use the interactive interface to always type the URL and select the authentication method manually.

I imagine something like this:
```bash
spacectl profile login --url https://myaccount.app.spacelift.io --method browser
```

varshneydevansh:
I am excited to contribute to this project, are there any issue which are a little beginner-friendly or any new feature implementation?

tomasmik:
> I am excited to contribute to this project, are there any issue which are a little beginner-friendly or any new feature implementation?

Hi! Sorry for the late reply, sadly contributing without having a spacelift account is not that simple as most features require at least a free version of spacelift while some require a paid one. Issues also aren't sorted and tagged or lack proper description for easy first time contributing.

PR #165, opened 2023-07-28
Allow to predefine login actions

tomasmik:
Added the ability to predefine the login path to take using flags, this allows for a quicker way to login with less clicking:
* You can define the `--endpoint` to skip the need to enter the spacelift url.
* You can define the `--method` to skip the choice of login method.
* You can define both the values above to skip both steps.

Also removed the confusing loop and updated selection for the login method to a prettier menu.

Command example:
```
spacectl profile login --method browser --endpoint https://unicorn.app.spacelift.io  local-test
```

Closes: https://github.com/spacelift-io/spacectl/issues/158

### DIFF
--- a/internal/cmd/profile/flags.go
+++ b/internal/cmd/profile/flags.go
@@ -1,6 +1,12 @@
 package profile
 
-import "github.com/urfave/cli/v2"
+import (
+	"fmt"
+
+	"github.com/urfave/cli/v2"
+
+	"github.com/spacelift-io/spacectl/client/session"
+)
 
 var bindHost string
 var flagBindHost = &cli.StringFlag{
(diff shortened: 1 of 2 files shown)


internal/cmd/profile/login_command.go
```diff
@@ -58,75 +60,89 @@ func loginAction(ctx *cli.Context) error {
...
+	case session.CredentialsTypeAPIKey:
+		if err := loginUsingAPIKey(reader, &storedCredentials); err != nil {
+			return err
+		}
+	case session.CredentialsTypeGitHubToken:
+		if err := loginUsingGitHubAccessToken(&storedCredentials); err != nil {
+			return err
+		}
+	case session.CredentialsTypeAPIToken:
+		return loginUsingWebBrowser(ctx, &storedCredentials)
```
mbialon, line 85: Unless an error occurs, other methods continue to `return persistAccessCredentials(&storedCredentials)`. Why do we exit here?

internal/cmd/profile/flags.go
```diff
@@ -21,3 +27,41 @@ var flagBindPort = &cli.IntFlag{
...
 	EnvVars:     []string{"SPACECTL_BIND_PORT"},
 }
+
+const (
+	methodBrowser = "browser"
+	methodAPI     = "api"
+	methodGithub  = "github"
+)
+
+var methodToCrendentialsType = map[string]session.CredentialsType{
```
mbialon, line 37:
```suggestion
var methodToCredentialsType = map[string]session.CredentialsType{
```

internal/cmd/profile/login_command.go
tomasmik: This one is a bit complicated as it opens a browser, starts a server etc. so it parses and saves the credentials in there as well. I don' think it's worth to refactor as it would make it more DRY but also more confusing.

mbialon (approved):

PR #171, opened 2023-08-04
Allow to get latest logs for a stack

tomasmik:
Instead of looking up all runs and picking the latest one, allow to provide a flag to query the latest run for the user and just get those logs. This is the most common use case for the command anyway.

### DIFF
--- a/internal/cmd/stack/flags.go
+++ b/internal/cmd/stack/flags.go
@@ -52,7 +52,12 @@ var flagRequiredRun = &cli.StringFlag{
 
 var flagRun = &cli.StringFlag{
 	Name:  "run",
-	Usage: "`ID` of the run",
+	Usage: "[Optional] `ID` of the run",
+}
+
+var flagRunLatest = &cli.BoolFlag{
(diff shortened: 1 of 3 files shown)


internal/cmd/stack/flags.go
```diff
@@ -52,7 +52,12 @@ var flagRequiredRun = &cli.StringFlag{
 
 var flagRun = &cli.StringFlag{
 	Name:  "run",
-	Usage: "`ID` of the run",
+	Usage: "[Optional] `ID` of the run",
+}
+
+var flagRunLatest = &cli.BoolFlag{
```
tomasmik, line 58: We could instead do this automatically if no `--run` was provided, but I feel like that's kinda unexpected. I'm torn here, so suggestions appreciated.

Yantrio (approved):

internal/cmd/stack/flags.go
Yantrio: I think for now, let's keep this as an option to explicitly set 👍

PR #177, opened 2023-09-04
Use filepath functions for path manipulation in getGitRepositorySubdir

JeffreyVdb:
Before this commit, the root directory would be removed which leaves a trailing file separator character in the beginning. This would then be trimmed by using strings.TrimPrefix. The problem here is that it uses a hardcoded file separator character which won't work on Windows because Windows uses a backslash instead of a forward slash.

Using filepath.Rel to get the directory relative to the root directory will not leave a separator character in the beginning.

It's also important to use filepath.ToSlash which will translate every file separator character to the forward slash.

This commit makes discovery work on Windows which it didn't before.

### DIFF
--- a/internal/cmd/stack/open_command.go
+++ b/internal/cmd/stack/open_command.go
@@ -122,6 +122,13 @@ func getGitRepositorySubdir() (string, error) {
 		} else if !os.IsNotExist(err) {
 			return "", fmt.Errorf("couldn't stat .git directory: %w", err)
 		}
+
+		if newRoot := filepath.Dir(root); newRoot != root {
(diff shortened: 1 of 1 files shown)


tomasmik:
Hey, thanks for your contribution. Please allow for some time to test this as this is Windows specific.

JeffreyVdb:
The last commit should also be applicable to Linux. It should endlessly loop on all platforms (before this fix).

tomasmik (approved):

tomasmik:
Tested and merged, thanks @JeffreyVdb

PR #182, opened 2023-09-27
Improve user facing error massage when the stack is not found

mwasilew2:
closes: https://app.clickup.com/t/862k09kgc

Includes in the error message a hint about how to list available stacks.

Note: The GraphQL API doesn't return errors that can be typed. The string parsing is a dirty hack, if anyone knows of a better way to approach it let me know.

### DIFF
--- a/internal/cmd/stack/stack_selector.go
+++ b/internal/cmd/stack/stack_selector.go
@@ -7,10 +7,20 @@ import (
 
 	"github.com/manifoldco/promptui"
 	"github.com/pkg/errors"
+	"github.com/shurcooL/graphql"
+	"github.com/spacelift-io/spacectl/internal/cmd/authenticated"
 	"github.com/urfave/cli/v2"
 )
 
 var errNoStackFound = errors.New("no stack found")
+
+type errStackWithIdNotFound struct {
+	stackId string
+}
+
+func (e errStackWithIdNotFound) Error() string {
+	return fmt.Sprintf("Stack with id %q could not be found. Please check that the stack exists and that you have access to it. To list available stacks run: spacectl stack list", e.stackId)
+}
 
 // getStackID will try to retreive a stack ID from multiple sources.
 // It will do so in the following order:
@@ -18,7 +28,12 @@ var errNoStackFound = errors.New("no stack found")
 // 2. Check the current directory to determine repository and subdirectory and search for a stack.
 func getStackID(cliCtx *cli.Context) (string, error) {
 	if cliCtx.IsSet(flagStackID.Name) {
-		return cliCtx.String(flagStackID.Name), nil
+		stackId := cliCtx.String(flagStackID.Name)
+		err := stackExists(cliCtx.Context, stackId)
+		if err != nil {
+			return "", err
+		}
+		return stackId, nil
 	}
 
 	subdir, err := getGitRepositorySubdir()
@@ -45,6 +60,29 @@ func getStackID(cliCtx *cli.Context) (string, error) {
 	}
 
 	return got, nil
+}
+
+func stackExists(ctx context.Context, stackId string) error {
+	var query struct {
+		Stack struct {
+			ID string `graphql:"id"`
+		} `graphql:"stack(id: $id)"`
+	}
+
+	variables := map[string]interface{}{
+		"id": graphql.ID(stackId),
+	}
+
+	err := authenticated.Client.Query(ctx, &query, variables)
+	if err != nil {
+		return fmt.Errorf("failed to query GraphQL API: %w", err)
+	}
+
+	if query.Stack.ID == "" {
+		return errStackWithIdNotFound{stackId: stackId}
+	}
+
+	return nil
 }
 
 func findAndSelectStack(ctx context.Context, p *stackSearchParams, forcePrompt bool) (string, error) {
--- a/internal/cmd/stack/stack_selector.go
+++ b/internal/cmd/stack/stack_selector.go
@@ -75,7 +75,7 @@ func stackExists(ctx context.Context, stackId string) error {
 
 	err := authenticated.Client.Query(ctx, &query, variables)
 	if err != nil {
-		return fmt.Errorf("failed to query GraphQL API: %w", err)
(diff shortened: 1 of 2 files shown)


Clickup-user:
Task linked: [CU-862k09kgc Add suggestions when spacectl does not find an entity](https://app.clickup.com/t/862k09kgc)

tomasmik:
We have this function called `getStackID` which is called every time you do anything with a command which interacts with stacks: https://github.com/spacelift-io/spacectl/blob/main/internal/cmd/stack/stack_selector.go#L19

Because this is called every time we parse the provided `--id` I was thinking maybe we could actually just look up and check if the provided stack actually exists in spacelift instead of just returning the provided name?

```go
	if cliCtx.IsSet(flagStackID.Name) {
		return cliCtx.String(flagStackID.Name), nil
	}
```

Would you mind look in to that and seeing if that actually works?

mwasilew2:
> We have this function called `getStackID` which is called every time you do anything with a command which interacts with stacks: https://github.com/spacelift-io/spacectl/blob/main/internal/cmd/stack/stack_selector.go#L19
> 
> Because this is called every time we parse the provided `--id` I was thinking maybe we could actually just look up and check if the provided stack actually exists in spacelift instead of just returning the provided name?
> 
> ```go
> 	if cliCtx.IsSet(flagStackID.Name) {
> 		return cliCtx.String(flagStackID.Name), nil
> 	}
> ```
> 
> Would you mind look in to that and seeing if that actually works?

You're right, it makes more sense to check for stack presence when we're extracting the stack id. The additional benefit of this is that this function is used in other places so we get nice error handling in those places as well.

@tomasmik  let me know what you think of this approach

internal/cmd/stack/stack_selector.go
```diff
@@ -7,18 +7,33 @@ import (
...
 	"github.com/manifoldco/promptui"
 	"github.com/pkg/errors"
+	"github.com/shurcooL/graphql"
+	"github.com/spacelift-io/spacectl/internal/cmd/authenticated"
 	"github.com/urfave/cli/v2"
 )
 
 var errNoStackFound = errors.New("no stack found")
 
+type errStackWithIdNotFound struct {
```
tomasmik, line 17: I would be for removing this but if you dont:

```suggestion
type errStackWithIDNotFound struct {
```

internal/cmd/stack/stack_selector.go
```diff
@@ -47,6 +62,29 @@ func getStackID(cliCtx *cli.Context) (string, error) {
...
+		} `graphql:"stack(id: $id)"`
+	}
+
+	variables := map[string]interface{}{
+		"id": graphql.ID(stackId),
+	}
+
+	err := authenticated.Client.Query(ctx, &query, variables)
+	if err != nil {
+		return fmt.Errorf("failed to query GraphQL API: %w", err)
```
tomasmik, line 78: Let's be more descriptive here:
```suggestion
		return fmt.Errorf("failed to query GraphQL API when checking if a stack exists: %w", err)
```

internal/cmd/stack/stack_selector.go
```diff
@@ -47,6 +62,29 @@ func getStackID(cliCtx *cli.Context) (string, error) {
 	return got, nil
 }
 
+func stackExists(ctx context.Context, stackId string) error {
```
tomasmik, line 65: Honestly I would change this to:
```suggestion
func stackExists(ctx context.Context, stackId string) (bool, error) {
```

Allowing the caller to choose what he wants to do with the information that the stacks exists or not. We could also get rid of the custom error then, as the caller could craft the error message itself.

internal/cmd/stack/stack_selector.go
```diff
@@ -7,18 +7,33 @@ import (
 
 	"github.com/manifoldco/promptui"
 	"github.com/pkg/errors"
+	"github.com/shurcooL/graphql"
+	"github.com/spacelift-io/spacectl/internal/cmd/authenticated"
```
tomasmik, line 11: Sorted imports are nice imports: https://github.com/spacelift-io/spacectl/blob/main/internal/cmd/run_external_dependency/mark_completed.go#L3

internal/cmd/stack/stack_selector.go
mwasilew2: got rid of it in the latest version

internal/cmd/stack/stack_selector.go
mwasilew2: yup, agreed

internal/cmd/stack/stack_selector.go
```diff
@@ -47,6 +57,28 @@ func getStackID(cliCtx *cli.Context) (string, error) {
...
+	variables := map[string]interface{}{
+		"id": graphql.ID(stackId),
+	}
+
+	err := authenticated.Client.Query(ctx, &query, variables)
+	if err != nil {
+		return false, fmt.Errorf("failed to query GraphQL API when checking if a stack exists: %w", err)
+	}
+
+	if query.Stack.ID == "" {
```
tomasmik, line 76: total nit:

```suggestion
	return query.Stack.ID != ""
```

tomasmik (approved):

internal/cmd/stack/stack_selector.go
mwasilew2: What do you mean by that? I've got autoformatting and linting enabled in my editor and these imports seem to be in lexicographic order.

tomasmik: That's fine, but not quite right. The default accepted way of ordering imports is:
```
STD library

external dependencies

internal dependencies
```

I think [goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports) should do that by default actually.

mwasilew2: ah, thx

for future reference, it was just a matter of toggling an option in code style settings:

Screenshot from 2023-09-29 11-38-17

tomasmik (approved):
Amazing

PR #185, opened 2023-09-28
Add installation guide for alpine linux package

hristiy4n:
I've added an alpine package for `spacectl` and I would like to update the `README.md` with a similar note like the arch linux package, in case others want to use it. 

The code for the package can be found [here](https://gitlab.alpinelinux.org/alpine/aports/-/tree/master/testing/spacectl) in case any changes need to be made


tomasmik (approved):
Hey, thanks for the contribution. Ideally we would like to have the readme structured a bit differently as we have a few community maintained packages now, but this will do. Thanks again!

PR #190, opened 2023-10-06
Allow to draw a basic worker pools view

tomasmik:
This is fairly basic though easily extendable, this allows you to:
* Browse a list of runs for a worker pool
* Open a run in a browser

It will automatically update the list every 5 seconds and the list is sorted by position. Terminal size detection could also be improved.

Would be nice to add filtering and custom sorting in the future.

Here is in action:
* 1 run, 1 dependency:

https://github.com/spacelift-io/spacectl/assets/40318863/4ee713f2-542f-4546-9dca-79d5ef43917f


Yantrio (approved):
LGTM 👍 nice usage of bubbletea's tables.

PR #201, opened 2023-11-17
Add space details to the stack object

tomasmik:


cube2222 (approved):

internal/cmd/stack/list.go
```diff
@@ -132,7 +132,13 @@ type stack struct {
 	State               string   `graphql:"state" json:"state,omitempty"`
 	StateSetAt          int64    `graphql:"stateSetAt" json:"stateSetAt,omitempty"`
 	TerraformVersion    string   `graphql:"terraformVersion" json:"terraformVersion,omitempty"`
-	TrackedCommit       struct {
+	SpaceDetails        struct {
+		ID          string  `graphql:"id" json:"id,omitempty"`
+		Name        string  `graphql:"name" json:"name,omitempty"`
+		Description string  `graphql:"description" json:"description,omitempty"`
+		ParentSpace *string `graphql:"parentSpace" json:"parentSpace,omitempty"`
```
mbialon, line 139: In `SpaceDetails` we have both `ID` and `Name`. Is `ParentSpace` the id or name?

mbialon, line 139: Could we resolve `ParentSpace` as a `SpaceDetails` struct?
In case someone has built deep space hierarchies.

mbialon (approved):

internal/cmd/stack/list.go
tomasmik: Not a thing by default, requires additional queries. By default it's just a string value, ID.

PR #202, opened 2023-11-22
chore: added install method with aqua

ponkio-o:
Hi, I added install method with [aqua](https://aquaproj.github.io/) in README.

related PR: https://github.com/aquaproj/aqua-registry/pull/16197

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,14 @@ asdf install spacectl latest
 asdf global spacectl latest
 ```
 
+#### aqua
+
+You can install this tool with [aqua](https://aquaproj.github.io/).
+
+```
+aqua g -i spacelift-io/spacectl
+```
+
 #### GitHub Release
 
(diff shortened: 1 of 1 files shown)


mbialon (dismissed):
LGTM

tomasmik (changes requested):
Thank you for contributing but actually I'm unsure about this. Never heard of aqua, need to check what it is.

tomasmik:
Sorry but unfortunately we've decided to decline this change. We feel that this package (version) manager is a bit too little known to be included in the readme in this way. However we've noted that `spacectl` can be installed using `aqua` and will include it when we update the readme to accommodate additional install instructions without obscuring others.

Thank you again for your contribution.

booniepepper:
What is the criteria for "known enough"? Aqua is a fantastic project, and it's been growing in popularity with Japanese companies. I expect it to continue growing in popularity over time and hit some kind of inflection point. There are a few listed here https://github.com/aquaproj/aqua#who-uses-aqua and also Mercari is using it now too

PR #209, opened 2023-12-22
Install shell completions

### LINKED ISSUE #71: Support bash completion
It would be great if spacectl had support for bash completion (https://cli.urfave.org/v2/#bash-completion)

Ideally this would be extended with custom functions to also allow for completion of parameters like `--id` (aka slug)

mbialon:
- generates shell completions while preparing the release
- includes shell completions in archives
- adds completion installation to brew formulae

Fixes #71

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ build
 dist
 key.*
 *.asc
+
+completions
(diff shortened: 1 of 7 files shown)


tomasmik (approved):
awesome

PR #214, opened 2024-01-24
feat(usage-view): download usage data in csv format

mwasilew2:
related to: https://app.clickup.com/t/8693k73r6

Adds support for downloading usage view data in csv format.

### DIFF
--- a/internal/cmd/profile/usage_view_csv_command.go
+++ b/internal/cmd/profile/usage_view_csv_command.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"fmt"
 	"io"
-	"log"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -15,16 +14,9 @@ import (
 	"github.com/spacelift-io/spacectl/internal/cmd/authenticated"
 )
 
-type queryArgs struct {
-	since   string
-	until   string
-	aspect  string
-	groupBy string
-}
-
 func usageViewCSVCommand() *cli.Command {
 	return &cli.Command{
-		Name:      "usage-view-csv",
+		Name:      "usage-csv",
 		Usage:     "Prints CSV with usage data for the current account",
 		ArgsUsage: cmd.EmptyArgsUsage,
 		Flags: []cli.Flag{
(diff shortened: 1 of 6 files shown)


internal/cmd/profile/usage_view_csv_command.go
```diff
@@ -0,0 +1,108 @@
...
+				return err
+			}
+			defer resp.Body.Close()
+			if resp.StatusCode != http.StatusOK {
+				return fmt.Errorf("unexpected status code: %d", resp.StatusCode)
+			}
+
+			// save response to a file
+			var filePath string
+			if !ctx.IsSet(flagUsageViewCSVFile.Name) {
```
tomasmik, line 70: If this is not set, we could just print to stdout. My reasoning is one could just call it like `spacectl profile usage-view-csv > myfile.csv`, it would also be possible to do additional modifications on the fly if this was possible.

One thing that makes this not great is that we also print things like `log.Println("Querying Spacelift for usage data...")
`

internal/cmd/profile/usage_view_csv_command.go
```diff
@@ -0,0 +1,108 @@
...
+				return fmt.Errorf("failed to create an HTTP request: %w", err)
+			}
+			q := req.URL.Query()
+			for k, v := range params {
+				q.Add(k, v)
+			}
+			req.URL.RawQuery = q.Encode()
+
+			// execute http query
+			log.Println("Querying Spacelift for usage data...")
```
tomasmik, line 58: we actually use `fmt.Println` in spacectl

internal/cmd/profile/usage_view_csv_command.go
```diff
@@ -0,0 +1,108 @@
...
+		Before: authenticated.Ensure,
+		Action: func(ctx *cli.Context) error {
+			// prep http query
+			args := &queryArgs{
+				since:   ctx.String(flagUsageViewCSVSince.Name),
+				until:   ctx.String(flagUsageViewCSVUntil.Name),
+				aspect:  ctx.String(flagUsageViewCSVAspect.Name),
+				groupBy: ctx.String(flagUsageViewCSVGroupBy.Name),
+			}
+			params := buildQueryParams(args)
```
tomasmik, line 46: Is `queryArgs` struct really needed? You could just pass the `ctx *cli.Context` and parse it there

tomasmik (approved):

internal/cmd/profile/usage_view_csv_command.go
mwasilew2: I guess a global logging flag would be helpful here. Since we don't have that, what we can do is write csv file to `stdout` and print messages to `stderr`. Although that would be inconsistent with how it's done in other parts of the project. Let me know what you think.

tomasmik: > I guess a global logging flag would be helpful here.

I guess, we should maybe create an issue for this.

> Let me know what you think.

I had the idea myself, but didnt suggest it because it is inconsistent. Honestly I have no strong feeling for this. You can merge the code as is.

PR #221, opened 2024-02-29
Add git to container

### LINKED ISSUE #220: The git command does not seem to exist in the spacectl container image
The `spacectl stack local-preview` command uses the git command at some point. It first searches for the .git directory in parent directories until it determines the base directory. After that it calls git, I believe here: https://github.com/spacelift-io/spacectl/blob/94ba6a0370be46723a6a661ffd1492699a8b0796/internal/cmd/stack/open_command.go#L78

The problem is that git isn't installed in the container which prompted me to do this:

```Dockerfile
FROM ghcr.io/spacelift-io/spacectl

RUN set -xe \
  && apk add --no-cache git
```

I need a wrapper command to do exactly what the binary does:

```bash
#!/bin/bash
set -euo pipefail

root_dir=$(git rev-parse --show-toplevel)
relative_dir=$(realpath --relative-to="$root_dir" "$(pwd)")

podman run --rm -it \
        --workdir /src/"${relative_dir}" \
        -v ${root_dir}:/src:Z \
        --userns=keep-id \
        --env=SPACELIFT_API_KEY_ENDPOINT \
        --env=SPACELIFT_API_KEY_ID \
        --env=SPACELIFT_API_KEY_SECRET \
        spacectl "$@"
```

This script basically mounts the git repository to /src and also moves you into the working directory you're executing the command from. This works using the container that I built.

JeffreyVdb:
Fixes #220

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,7 @@
 FROM alpine
+
+RUN set -xe \
+    && apk add --no-cache git
+
 COPY spacectl /usr/local/bin/spacectl
 ENTRYPOINT ["/usr/local/bin/spacectl"]


tomasmik:
Thank you for the contribution!

tomasmik (approved):

PR #227, opened 2024-05-03
Add spacelift client type request header

michieldewilde:


sephriot (approved):

eliecharra:
@michieldewilde I was off next week so maybe I missed it, but is there a central place where all the clients can be viewed and documented?

Also, let us know if we need to add this header to the calls made from k8s controller 🙏🏻

PR #231, opened 2024-06-26
Bump Go packages

peterdeme:
Just some chores.


.github/workflows/lint.yml
```diff
@@ -12,7 +12,6 @@ jobs:
   lint:
     name: Lint the code
     runs-on: ubuntu-latest
-    container: golang:1.22
```
peterdeme, line 15: no particular reason, I just wanted less place to maintain the version at. this way, it's only defined in `go.mod` file and nowhere else. 😊

.goreleaser.yml
```diff
@@ -1,3 +1,5 @@
+version: 2
```
peterdeme, line 1: this is mandatory for the new goreleaser version, otherwise it throws an error

eliecharra (approved):

Issue #236, opened 2024-07-19
mTLS connections to Spacelift API Endpoint


logand22:
There might be a case in a self-hosted environment in which the Spacelift graphQL endpoint is behind an mTLS configured proxy. If possible, consider adding additional configuration options in the `spacectl` binary in order to configure client keys and certs as well as a CA bundle.

PR #237, opened 2024-07-19
Allow for mTLS connections configured from environment

logand22:
Created an example PR for issue: https://github.com/spacelift-io/spacectl/issues/236 in which you can configure Client TLS settings through environment variables. Let me know if this is contribution you would be willing to accept or if I should request this feature through other channels. 

Additionally, there might be improved UX if the client certs could be linked to a particular profile as well although I didn't explore that in this PR.


peterdeme:
thanks for the contribution! can you fix the warning please? otherwise lgtm.

logand22:
> thanks for the contribution! can you fix the warning please? otherwise lgtm.

Fixed!

tomasmik (approved):

PR #239, opened 2024-08-06
API Endpoint from Session Endpoint

logand22:
This PR modifies the `FromAPIToken` function to accept an additional endpoint argument. This ensures that the Spacelift session manager respects the profile endpoint. We also don't change the behavior of loading the API token from the environment and in that case we respect the original intentions. 

## Alternatives
I could remove the following logic, however this would change the behavior of the environmental variables such that SPACELIFT_API_ENDPOINT or SPACELIFT_API_KEY_ENDPOINT are required which seems like a breaking change. 
```go
if endpoint != "" {
```

### DIFF
--- a/client/session/api_token.go
+++ b/client/session/api_token.go
@@ -18,8 +18,8 @@ import (
 const timePadding = 30 * time.Second
 
 // FromAPIToken creates a session from a ready API token.
-func FromAPIToken(_ context.Context, client *http.Client) func(string) (Session, error) {
-	return func(token string) (Session, error) {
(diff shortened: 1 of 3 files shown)


tomasmik:
Hi, thanks for the contribution! I've tagged reviewers, we'll get this reviewed as soon as possible.

peterdeme (approved):

mbialon (approved):

logand22:
Bump. Looks like it needs to be merged.

Issue #241, opened 2024-08-21
Allow to prioritise runs 


tomasmik:
We allow users in the [UI to prioritise runs](https://docs.spacelift.io/concepts/run.html#run-prioritization) we should also allow that in spacectl. To do that we should add an additional sub-command on `spacectl stack` which would prioritise a run.

Bonus: Add a flag which would prioritise runs created by the `local-preview` command.

PR #243, opened 2024-08-26
feat: Add new commands for prioritizing/deprioritizing a run and flag to prioritize local-preview

0michalsokolowski0:
Closes https://github.com/spacelift-io/spacectl/issues/241.

In this PR, I am:
* Adding a new command `prioritize` to the stack allows you to prioritize a run to match the functionality in the UI.
Example:
``` bash 
spacectl stack prioritize --id=<stack-id> --run=<run-id>
```
image

* Adding a new command `deprioritize` to the stack allows you to deprioritize a run to match the functionality in the UI.
Example:
``` bash 
spacectl stack deprioritize --id=<stack-id> --run=<run-id>
```

* Adding a flag `prioritize-run` to local-preview, which allows to prioritize the local-preview run after it's triggered automatically
Example:
``` bash
spacectl stack local-preview --id=<stack-id> --prioritize-run
```
image


mbialon (approved):
Looks good.

internal/cmd/stack/local_preview.go
```diff
@@ -119,12 +119,27 @@ func localPreview() cli.ActionFunc {
...
+		if cliCtx.Bool(flagPrioritizeRun.Name) {
+			var prioritizeMutation setRunPriorityMutation
+			variables := map[string]interface{}{
+				"stackId":    graphql.ID(stack.ID),
+				"runId":      graphql.ID(triggerMutation.RunProposeLocalWorkspace.ID),
+				"prioritize": graphql.Boolean(true),
+			}
+
+			if err = authenticated.Client.Mutate(ctx, &prioritizeMutation, variables); err != nil {
+				return err
```
tomasmik, line 137: We should probably print the error and advice the caller to do this himself instead of failing outright. The run itself got created just fine, no reason to exit.

internal/cmd/stack/run_prioritize.go
```diff
@@ -0,0 +1,52 @@
...
+		return err
+	}
+	runID := cliCtx.String(flagRequiredRun.Name)
+
+	var mutation setRunPriorityMutation
+
+	variables := map[string]interface{}{
+		"stackId":    graphql.ID(stackID),
+		"runId":      graphql.ID(runID),
+		"prioritize": graphql.Boolean(true),
```
tomasmik, line 28: Should we extend this so that the caller could set this? This customer don't have to go to UI to revert their action in case they make a mistake.

internal/cmd/stack/run_prioritize.go
```diff
@@ -0,0 +1,52 @@
+package stack
+
+import (
+	"fmt"
```
tomasmik, line 4: [imports not sorted?](https://github.com/spacelift-io/spacectl/blob/89a33d36e07ed88308b1fac96ab1dbdfd90886d3/internal/cmd/stack/local_preview.go#L3)

tomasmik (commented):
Nice work, just a few little nits

internal/cmd/stack/run_prioritize.go
0michalsokolowski0: Maybe I should add the golangci-linter config to detect it in ci?

internal/cmd/stack/run_prioritize.go
0michalsokolowski0: In my opinion, from the user's perspective, it would be easier to have a separate command "deprioritize". What do you think?

0michalsokolowski0: f6fcd7349259762110b8a9782db8683fda2e3b8b

internal/cmd/stack/local_preview.go
0michalsokolowski0: f6fcd7349259762110b8a9782db8683fda2e3b8b

internal/cmd/stack/run_prioritize.go
0michalsokolowski0: https://github.com/spacelift-io/spacectl/pull/244

tomasmik (approved):
good work

PR #246, opened 2024-08-28
refactor: Use searchStacks over stacks for listing stacks

0michalsokolowski0:


internal/cmd/stack/list.go
```diff
@@ -23,57 +20,40 @@ func listStacks() cli.ActionFunc {
...
 
 		return fmt.Errorf("unknown output format: %v", outputFormat)
 	}
 }
 
-func listStacksJSON(ctx context.Context) error {
-	var query struct {
-		Stacks []stack `graphql:"stacks" json:"stacks,omitempty"`
+func listStacksJSON(ctx *cli.Context) error {
+	stacks, err := searchStacks(ctx.Context, structs.SearchInput{})
```
0michalsokolowski0, line 31: The order here is not the same as previously. Does that matter?

internal/cmd/stack/list.go
```diff
@@ -23,57 +20,40 @@ func listStacks() cli.ActionFunc {
...
-	}
-
-	if err := authenticated.Client.Query(ctx.Context, &query, map[string]interface{}{}); err != nil {
-		return errors.Wrap(err, "failed to query list of stacks")
-	}
-
-	sort.SliceStable(query.Stacks, func(i, j int) bool {
-		return strings.Compare(strings.ToLower(query.Stacks[i].Name), strings.ToLower(query.Stacks[j].Name)) < 0
+	stacks, err := searchStacks(ctx.Context, structs.SearchInput{
+		OrderBy: &structs.QueryOrder{
```
tomasmik, line 41: I wonder if we should follow the same logic as the `default view` in the UI for stacks? I think that one is
```
"orderBy": [
    {
        "field": "starred",
        "direction": "DESC"
    }
]
``

internal/cmd/stack/list.go
tomasmik: Probably no.

internal/cmd/stack/list.go
0michalsokolowski0: Yes, precisely. 👍🏻

0michalsokolowski0: 2d7797ce4b02f3000d234c94d1aa2e0f38c03eb2

tomasmik (approved):

Issue #250, opened 2024-08-29
Local preview does not exit on run failure


tomasmik:
When a run enters failed state, it will hang and not exit while it will exit all other states.

beforster:
@tomasmik Thanks for fixing this. My team is affected by this bug also. Could we get this fix released soon? If not, do you know when this bug was introduced so that we can rollback to a prior version?

PR #251, opened 2024-08-29
fix: Watch log hangs when we're in a terminal state

tomasmik:
If we have no more logs and we're in a terminal state, don't wait for more, just exit. This fixes the problem of local preview not exiting.

```
....
╷
│ Error: Unsupported block type
│
│   on main.tf line 1:
│    1: aarsource "random_string" "random" {
│
│ Blocks of type "aarsource" are not expected here.
╵

[01J6F1AVF3VFZ7NZ7K4EMEKMW6] Unexpected exit code when initializing workspace: 1

-----------------
FAILED	Thu Aug 29 15:35:31 EEST 2024
-----------------

View full logs at http://*.app.spacelift.tf/stack/doublestack/run/01J6F2462RPSAPFCRY6GMP2FPT
2024/08/29 15:35:32 finished with FAILED state
➜  $

```

This was reported by a customer.

Closes: https://github.com/spacelift-io/spacectl/issues/250

### DIFF
--- a/internal/cmd/stack/run_logs.go
+++ b/internal/cmd/stack/run_logs.go
@@ -129,7 +129,7 @@ func runStates(ctx context.Context, stack, run string, sink chan<- string, acFn 
 			fmt.Println("")
 
 			if transition.HasLogs {
-				if err := runStateLogs(ctx, stack, run, transition.State, transition.StateVersion, sink); err != nil {
+				if err := runStateLogs(ctx, stack, run, transition.State, transition.StateVersion, sink, transition.Terminal); err != nil {
 					return nil, err
 				}
 			}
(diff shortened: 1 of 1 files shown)


internal/cmd/stack/run_logs.go
```diff
@@ -205,7 +205,7 @@ func runStateLogs(ctx context.Context, stack, run string, state structs.RunState
 			sink <- message.Body
 		}
 
-		if logs.Finished {
+		if logs.Finished || !logs.HasMore && stateTerminal {
```
mbialon, line 208:
```suggestion
		if logs.Finished || (!logs.HasMore && stateTerminal) {
```
Is this the same? So I don't have to think about operator priority.

mbialon (commented):
Is the goal to exit when we fetch all logs for a terminal transition? Because we can't rely on the `Finished` flag.

mbialon (approved):

PR #252, opened 2024-08-29
feat: Add limit and search flags to stack list command

0michalsokolowski0:
Adds limit and search flags to stack list command

### DIFF
--- a/internal/cmd/stack/list.go
+++ b/internal/cmd/stack/list.go
@@ -31,7 +31,7 @@ func listStacks() cli.ActionFunc {
 				return fmt.Errorf("limit must be less than %d", math.MaxInt32)
 			}
 
-			limit = internal.ToPtr(cliCtx.Uint(flagLimit.Name))
+			limit = internal.Ptr(cliCtx.Uint(flagLimit.Name))
 		}
 
 		var search *string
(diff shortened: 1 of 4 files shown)


internal/helpers.go
```diff
@@ -0,0 +1,5 @@
+package internal
+
+func ToPtr[T any](v T) *T {
```
mbialon, line 3: Nit: `Ptr` would be fine.

mbialon (approved):
LGTM

internal/cmd/stack/list.go
```diff
@@ -18,20 +21,57 @@ func listStacks() cli.ActionFunc {
 			return err
 		}
 
+		var limit *uint
+		if cliCtx.IsSet(flagLimit.Name) {
+			if cliCtx.Uint(flagLimit.Name) == 0 {
+				return fmt.Errorf("limit must be greater than 0")
+			}
+
+			if cliCtx.Uint(flagLimit.Name) >= math.MaxInt32 {
```
tomasmik, line 30: Are such checks really necessary? No point in removing it now, but also eeh 🤷🏻

Yantrio: I like it :) Validation is helpful

internal/helpers.go
0michalsokolowski0: 8d92c92027f44e5d3683a7d8d837dbc0cc29ebf9

internal/cmd/stack/list.go
0michalsokolowski0: It might be a matter of taste, but I like knowing what to expect from inputs  :)

Issue #230, opened 2024-05-21
Find Stack given Run ID


Nuru:
In many places, particularly logging output captured by Datadog, I have access to a Run ID, but not the stack name. In general (maybe in all cases), `spacectl` and the web UI can do nothing with a Run ID unless you also supply the stack ID/slug.

Please add a way to get the Stack ID associated with a Run ID.

tomasmik:
While in theory this is possible, it's not as simple as just adding it to spacectl because backend doesn't provide a way to resolve a run (and to which stack it belongs) by just providing a run ID. To complete this, we would first need to extend backend.

PR #253, opened 2024-08-30
feat: Allow to provide run ID instead of stack ID in some command

0michalsokolowski0:
Closes https://github.com/spacelift-io/spacectl/issues/230

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -43,7 +43,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Run GoSec Security Scanner
-        uses: securego/gosec@master
+        uses: securego/gosec@v2.21.0
         with:
           args: "-severity=medium -no-fail -fmt sarif -out gosec-results.sarif ./..."
 
--- a/internal/cmd/stack/stack_selector.go
+++ b/internal/cmd/stack/stack_selector.go
@@ -18,6 +18,7 @@ var errNoStackFound = errors.New("no stack found")
 // getStackID will try to retrieve a stack ID from multiple sources.
 // It will do so in the following order:
 // 1. Check the --id flag, if set, use that value.
+// 2. Check the --run flag, if set, try to get the stack associated with the run.
 // 2. Check the current directory to determine repository and subdirectory and search for a stack.
 func getStackID(cliCtx *cli.Context) (string, error) {
 	stack, err := getStack(cliCtx)
@@ -39,6 +40,16 @@ func getStack(cliCtx *cli.Context) (*stack, error) {
 			return nil, fmt.Errorf("stack with id %q could not be found. Please check that the stack exists and that you have access to it. To list available stacks run: spacectl stack list", stackID)
 		}
 		return stack, nil
+	} else {
+		if cliCtx.IsSet(flagRun.Name) {
+			runID := cliCtx.String(flagRun.Name)
+			stack, err := stackGetByRunID(cliCtx.Context, runID)
+			if err == nil {
+				return stack, nil
+			}
+
+			// TODO: Maybe we should log error here?
+		}
 	}
 
 	subdir, err := getGitRepositorySubdir()
@@ -88,6 +99,25 @@ func stackGetByID(ctx context.Context, stackID string) (*stack, error) {
 	}
 
 	return &query.Stack.stack, nil
+}
+
+func stackGetByRunID(ctx context.Context, runID string) (*stack, error) {
+	var query struct {
+		StackByRunId struct {
+			stack
+		} `graphql:"stackByRunId(runId: $runId)"`
+	}
+
+	variables := map[string]interface{}{
+		"runId": graphql.ID(runID),
+	}
+
+	err := authenticated.Client.Query(ctx, &query, variables)
+	if err != nil {
+		return nil, fmt.Errorf("failed to query GraphQL API when getting stack by run id: %w", err)
+	}
+
+	return &query.StackByRunId.stack, nil
 }
 
 func findAndSelectStack(ctx context.Context, p *stackSearchParams, forcePrompt bool) (*stack, error) {
--- a/internal/cmd/stack/stack_selector.go
+++ b/internal/cmd/stack/stack_selector.go
@@ -46,13 +46,13 @@ func getStack(cliCtx *cli.Context) (*stack, error) {
 		return stack, nil
 	} else if cliCtx.IsSet(flagRun.Name) {
 		runID := cliCtx.String(flagRun.Name)
-		stack, found, err := stackGetByRunID(cliCtx.Context, runID)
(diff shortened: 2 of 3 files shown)


internal/cmd/stack/stack_selector.go
```diff
@@ -31,14 +34,28 @@ func getStackID(cliCtx *cli.Context) (string, error) {
 func getStack(cliCtx *cli.Context) (*stack, error) {
 	if cliCtx.IsSet(flagStackID.Name) {
 		stackID := cliCtx.String(flagStackID.Name)
-		stack, err := stackGetByID(cliCtx.Context, stackID)
+		stack, found, err := stackGetByID(cliCtx.Context, stackID)
 		if err != nil {
 			return nil, fmt.Errorf("failed to check if stack exists: %w", err)
 		}
-		if errors.Is(err, errNoStackFound) {
```
0michalsokolowski0, line 38: It doesn't work, replaced it with a bool variable

mbialon: We can't reach this condition with a non-nil error because we handle all errors on line 38.

We have to reorder the conditions to make it work, like this:

```
stack, err := stackGetByID(cliCtx.Context, stackID)
if errors.Is(err, errNoStackFound) {
	return nil, fmt.Errorf("stack with id %q could not be found. Please check that the stack exists and that you have access to it. To list available stacks run: spacectl stack list", stackID)
}
if err != nil {
	return nil, fmt.Errorf("failed to check if stack exists: %w", err)
}
```

mbialon (changes requested):

internal/cmd/stack/stack_selector.go
0michalsokolowski0: True, I didn't spot that. Thanks!

internal/cmd/stack/stack_selector.go
```diff
@@ -31,14 +34,28 @@ func getStackID(cliCtx *cli.Context) (string, error) {
...
+		if !found {
 			return nil, fmt.Errorf("stack with id %q could not be found. Please check that the stack exists and that you have access to it. To list available stacks run: spacectl stack list", stackID)
 		}
+
 		return stack, nil
+	} else if cliCtx.IsSet(flagRun.Name) {
+		runID := cliCtx.String(flagRun.Name)
+		stack, found, err := stackGetByRunID(cliCtx.Context, runID)
+		if err != nil {
+			fmt.Printf("Failed to get stack by run id: %v\n", err)
```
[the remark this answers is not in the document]
0michalsokolowski0: I did it this way, assuming we want to try to resolve stack from run if possible optionally. However, on second thought, a user-provided invalid run ID that does not point to any particular stack suggests that it is incorrect. In that case, I should fail. Good point, thanks!

internal/cmd/stack/stack_selector.go
0michalsokolowski0: def7656745a4e1ffb4c1becb5ee18c02fb95ecee

internal/cmd/stack/stack_selector.go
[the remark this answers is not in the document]
0michalsokolowski0: def7656745a4e1ffb4c1becb5ee18c02fb95ecee

internal/cmd/stack/stack_selector.go
```diff
@@ -33,11 +36,25 @@ func getStack(cliCtx *cli.Context) (*stack, error) {
...
+
 			return nil, fmt.Errorf("failed to check if stack exists: %w", err)
 		}
-		if errors.Is(err, errNoStackFound) {
-			return nil, fmt.Errorf("stack with id %q could not be found. Please check that the stack exists and that you have access to it. To list available stacks run: spacectl stack list", stackID)
+
+		return stack, nil
+	} else if cliCtx.IsSet(flagRun.Name) {
+		runID := cliCtx.String(flagRun.Name)
+		stack, found, err := stackGetByRunID(cliCtx.Context, runID)
```
tomasmik, line 49: why the double logic? just return the same error as we do in the previous method maybe?

tomasmik (approved):

internal/cmd/stack/stack_selector.go
0michalsokolowski0: 41ffe56627bc02752102acb40d82a3305c29e7f7

internal/cmd/stack/stack_selector.go
```diff
@@ -34,28 +34,28 @@ func getStackID(cliCtx *cli.Context) (string, error) {
 func getStack(cliCtx *cli.Context) (*stack, error) {
 	if cliCtx.IsSet(flagStackID.Name) {
 		stackID := cliCtx.String(flagStackID.Name)
-		stack, found, err := stackGetByID(cliCtx.Context, stackID)
+		stack, err := stackGetByID(cliCtx.Context, stackID)
 		if err != nil {
-			return nil, fmt.Errorf("failed to check if stack exists: %w", err)
-		}
+			if errors.Is(err, errNoStackFound) {
```
mbialon, line 39: Nit: You don't have to check if the error exists to call `errors.Is`. I prefer less nested indentation.

internal/cmd/stack/stack_selector.go
```diff
@@ -34,28 +34,28 @@ func getStackID(cliCtx *cli.Context) (string, error) {
...
 		return stack, nil
 	} else if cliCtx.IsSet(flagRun.Name) {
 		runID := cliCtx.String(flagRun.Name)
-		stack, found, err := stackGetByRunID(cliCtx.Context, runID)
+		stack, err := stackGetByRunID(cliCtx.Context, runID)
 		if err != nil {
-			fmt.Printf("Failed to get stack by run id: %v\n", err)
-		} else {
-			if !found {
+			if errors.Is(err, errNoStackFound) {
```
mbialon, line 51: Nit: same as before.

mbialon (approved):
Good job!

internal/cmd/stack/stack_selector.go
0michalsokolowski0: 64b2af1071b8f7f8c47dbf91688069beeba079d5

internal/cmd/stack/stack_selector.go
0michalsokolowski0: 64b2af1071b8f7f8c47dbf91688069beeba079d5

.github/workflows/security.yml
```diff
@@ -43,7 +43,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Run GoSec Security Scanner
-        uses: securego/gosec@master
+        uses: securego/gosec@v2.21.0
```
0michalsokolowski0, line 46: I had to do that to fix the pipeline. The latest version produces an unsupported format which results in error **Unable to upload "gosec-results.sarif" as it is not valid SARIF**

Issue #254, opened 2024-09-13
Feature request: `spacectl module delete-version`


nessex:
Hi 👋 would it be possible to get a `spacectl module delete-version` command added to spacectl?

I'm currently setting up a workflow for our modules that involves automatically creating many versions. However, these versions are only useful for a short time. Given each version will sit around in the UI, and each will involve an archive getting created and pushed to e.g. S3, I'd really like to be able to clean up. The UI offers up a way to delete versions, but deleting them manually will be impractical in this case and I'd like to automate it. Having a command in the CLI would make this much easier!

tomasmik:
This is tricky, spacectl isn't meant for create or delete operations, but in this case maybe it makes sense.

nessex:
Thanks! I think there is a path we could take with the GraphQL API to self-serve on this regardless, but I figured it'd kind of fit alongside `spacectl module create-version`.

tomasmik:
@nessex your feature will be included in the next release, thanks for submitting a request :)

nessex:
Thanks so much!

PR #256, opened 2024-09-13
Allow to delete a module version

tomasmik:
```bash
➜  spc module delete-version --id gitlbam --versionid 01J537F77FRHBXVSEPXWNGFBX4
Module version "1.7.5" has been deleted
```

We already allow you to create versions using spacectl, so let's allow you to delete them as well.

Closes: https://github.com/spacelift-io/spacectl/issues/254

### DIFF
--- a/internal/cmd/module/flags.go
+++ b/internal/cmd/module/flags.go
@@ -8,6 +8,12 @@ var flagModuleID = &cli.StringFlag{
 	Required: true,
 }
 
+var flagVersionID = &cli.StringFlag{
+	Name:     "versionid",
+	Usage:    "[Required] User-facing `ID` (slug) of the version",
+	Required: true,
+}
+
 var flagCommitSHA = &cli.StringFlag{
 	Name:  "sha",
 	Usage: "Commit `SHA` to use for the module version",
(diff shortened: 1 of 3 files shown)


internal/cmd/module/flags.go
```diff
@@ -8,6 +8,12 @@ var flagModuleID = &cli.StringFlag{
 	Required: true,
 }
 
+var flagVersionID = &cli.StringFlag{
+	Name:     "versionid",
```
tomasmik, line 12: not a fan of this, but life is life. unless someone has a better idea?

0michalsokolowski0 (approved):

internal/cmd/module/flags.go
mbialon: Maybe we could provide the version itself, not the ID?

mbialon (approved):
LGTM left a comment about the version flag.

internal/cmd/module/flags.go
tomasmik: that would require changes in the backend, so i'd rather not do that :/

PR #264, opened 2024-09-24
Further improve repository string detection for open command

jeremyhager:
This further improves upon the existing repo name detection here: https://github.com/spacelift-io/spacectl/pull/233

While the current code works for `https://` and `git@`, it does not work for those who use a `ssh://git@` prefix:

example:
```shell
❯ spacectl stack open
2024/10/08 09:23:40 unsupported repository string: ssh://git@github.com/owner/repo.git
```

The [gh cli repo](https://github.com/cli/cli/blob/v2.57.0/git/url.go) already has existing logic for a handful of possible git urls, so this PR includes a dependency update so the `/cli/git` package can be imported and used for repo detection. 

Both `spacectl` and `cli` are MIT-licensed, so there (hopefully) won't be licensing concerns 
https://github.com/spacelift-io/spacectl/blob/main/LICENSE
https://github.com/cli/cli/blob/v2.57.0/LICENSE


tomasmik:
I will submit this as a separate PR as this has conflicts.

PR #267, opened 2024-11-07
Return an error when one then more alias is provided

tomasmik:
```
➜  spacectl git:(fix-alias) ✗ ./spc profile login asdf asdf
2024/11/07 11:07:40 expecting profile alias as the only argument, got 2 instead
➜  spacectl git:(fix-alias) ✗ ./spc profile login myorg
Enter Spacelift endpoint (eg. https://unicorn.app.spacelift.io/): https://tomasmik.app.spacelift.tf/
➜  spacectl git:(fix-alias) ✗ ./spc profile login
Waiting for login responses at 127.0.0.1:65014
```

### DIFF
--- a/internal/cmd/profile/get_alias.go
+++ b/internal/cmd/profile/get_alias.go
@@ -6,16 +6,20 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
-var (
-	profileAlias string
-)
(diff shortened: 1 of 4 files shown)


mbialon (approved):
LGTM

Issue #271, opened 2024-11-13
Make spacectl aware of values on auto-attached contexts


spr-mweber3:
I’ve a questing regarding auto-attached contexts. I was under the assumption, that the environment of a stack will be the merged view of all contexts attached, with priority in play and local overrides winning in each stack. This should include also variables/files on auto-attached contexts. But this doesn’t seem to be the case.

We’re using `spacectl` to actually fetch everything from the environment but are lacking the information on auto-attached contexts.

Would be nice if `spacectl` would return back the whole environment of a stack.

PR #274, opened 2024-11-27
feat: Include autoattach contexts when listing stack env

tomasmik:
This is a bit confusing now, but I think this is the easiest way to not accidentally duplicate the information returned.

```
Name                            | Type                 | Value      | Context | IsAutoattached
SPACELIFT_API_TOKEN             | ENVIRONMENT_VARIABLE | <computed> |         |
SPACELIFT_OIDC_TOKEN            | ENVIRONMENT_VARIABLE | <computed> |         |
TF_VAR_spacelift_account_name   | ENVIRONMENT_VARIABLE | <computed> |         |
...
bbbb                            | ENVIRONMENT_VARIABLE | bbb        | aaa     | false
vvv                             | ENVIRONMENT_VARIABLE | aaa        |         |
aaa                             | ENVIRONMENT_VARIABLE | aaa        | asdf111 | true
```

This is a result for a stack that has a context `bbbb` attached,  `aaa` auto-attached and `vvv` set as a normal variable.

### DIFF
--- a/internal/cmd/stack/environment.go
+++ b/internal/cmd/stack/environment.go
@@ -46,8 +46,11 @@ type listEnvElementOutput struct {
 	Name  string  `json:"name"`
 	Type  string  `json:"type"`
 	Value *string `json:"value"`
+
 	// Context specifies the name of the context.
-	Context *string `json:"context"`
+	Context        *string `json:"context"`
+	IsAutoattached *bool   `json:"isAutoattached"`
+
 	// Runtime is not printed, it's just used to determine output formatting.
 	Runtime bool `json:"runtime"`
 	// WriteOnly is not printed, it's just used to determine output formatting.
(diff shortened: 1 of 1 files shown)


internal/cmd/stack/environment.go
```diff
@@ -140,16 +150,37 @@ func (e *listEnvCommand) listEnv(cliCtx *cli.Context) error {
 	for _, config := range query.Stack.RuntimeConfig {
 		config := config
 		var contextName *string
+		var isAutoAttached *bool
 		if config.Context != nil {
 			contextName = &config.Context.ContextName
+
+			f := false
+			isAutoAttached = &f
```
mbialon, line 158:
```
isAutoAttached = &[]bool{f}[0]
```

I saw this "trick" yesterday, I don't yet know if I like it or not.

mbialon (approved):

spr-mweber3:
Fixes https://github.com/spacelift-io/spacectl/issues/271 ...

Thanks for that!

internal/cmd/stack/environment.go
tomasmik: I'm a man of few tricks and this will not be another one of them.

PR #276, opened 2024-12-06
Add policy commands

tomasmik:
This adds:

```
➜  spacectl git:(add-policy-operations) ✗ ./spc policy
NAME:
   spacectl policy - Manage Spacelift policies

USAGE:
   spacectl policy command [command options]

COMMANDS:
   list      List the policies you have access to
   show      Shows detailed information about a specific policy
   samples   List all policy samples
   sample    Inspect one policy sample
   simulate  Simulate a policy using a sample
```

All the commands are pretty basic and work as any other one. The simulate one is basic as well and can only simulate policies which already exist however the data provided can be custom, from a file or terminal input.

### DIFF
--- a/internal/cmd/policy/simulate.go
+++ b/internal/cmd/policy/simulate.go
@@ -1,0 +1,71 @@
+package policy
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/pkg/errors"
+	"github.com/shurcooL/graphql"
+	"github.com/spacelift-io/spacectl/internal/cmd"
+	"github.com/spacelift-io/spacectl/internal/cmd/authenticated"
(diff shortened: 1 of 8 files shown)


internal/cmd/policy/simulate.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	if err := authenticated.Client.Mutate(cliCtx.Context, &mutation, variables); err != nil {
+		return err
+	}
+
+	return cmd.OutputJSON(mutation.PolicySimulate)
+}
+
+func parseInput(input string) (string, error) {
```
silviutanasa, line 52: this is a little bit difficult to read

silviutanasa (approved):

PR #281, opened 2024-12-31
Replace shurcooL/graphql with hasura/go-graphql-client

### LINKED ISSUE #280: Use hasura/go-graphql-client to avoid doing replace in go.mod for users of the Go client
[https://github.com/hasura/go-graphql-client](https://github.com/hasura/go-graphql-client) is a maintained fork of [github.com/shurcooL/graphql](github.com/shurcooL/graphql), and it is also being used by other [spacelift code](https://github.com/spacelift-io/prometheus-exporter/pull/32/files).

To use the current client library, I need to add `replace github.com/shurcooL/graphql => github.com/spacelift-io/graphql v1.2.0` in my own repo.

RangelReale:
- replace shurcooL/graphql with hasura/go-graphql-client
- make `RequestOption` an internal option
- remove all deprecated `graphql.TYPE` with raw types

Fixes #280


tomasmik:
Thanks for the pull request. This might need extensive testing because there's a reason we use a custom graphql client. No promises or prioritising this.

Issue #279, opened 2024-12-20
Allow to explore the audit trail


tomasmik:
Add the [audit trail](https://docs.spacelift.io/integrations/audit-trail.html) command with sub-commands:
* `list`
* `show`

They should function in a similar fashion as any other search and show command does in spacectl, for example [blueprints](https://github.com/spacelift-io/spacectl/blob/main/internal/cmd/blueprint/list.go). For how to draw the table, look for inspiration in our frontend.

PR #282, opened 2025-01-15
feat: Introduce audit-trail subcommand

Axot017:
Partially closes: https://github.com/spacelift-io/spacectl/issues/279. Only `list` subcommand was added because graphql api does not expose for now resolver for querying single audit trail entry by id.

image


internal/cmd/audit_trail/list.go
```diff
@@ -0,0 +1,248 @@
...
+func listAuditTrails() cli.ActionFunc {
+	return func(cliCtx *cli.Context) error {
+		outputFormat, err := cmd.GetOutputFormat(cliCtx)
+		if err != nil {
+			return err
+		}
+
+		var limit *uint
+		if cliCtx.IsSet(cmd.FlagLimit.Name) {
+			if cliCtx.Uint(cmd.FlagLimit.Name) == 0 {
```
tomasmik, line 32: Just a thought, do you think these are useful? I think we have same checks in other list functions, but I'm just curious whats your opinion on checks like this.

main.go
```diff
@@ -5,6 +5,7 @@ import (
 	"os"
 	"time"
 
+	audittrail "github.com/spacelift-io/spacectl/internal/cmd/audit_trail"
```
tomasmik, line 8: could we rename the dir to `audittrail`? I'm not a fan of named imports :/

tomasmik (approved):
😎

main.go
Axot017: Sure, I will change that. I'm also not a fan of named imports. I did that only because this is how run-external-dependency command was done.

internal/cmd/audit_trail/list.go
Axot017: I think most of those checks are ok. I like showing reasonable error message when input has no sense. However I'm not a fan of this check:
```go
if cliCtx.Uint(cmd.FlagLimit.Name) == 0 {
	return fmt.Errorf("limit must be greater than 0")
}
```
and i added it only because we have it in other places. In my opinion `0` should be allowed value. We event say that `0` is default 
```
   --limit value               [Optional] Limit the number of items to return (default: 0)
```
so not allowing explicit `0` is a little bit weird to me

tomasmik: I'm cool with removing such checks

PR #283, opened 2025-01-17
feat: added stack sync-commit subcommand

sephriot:


tomasmik (approved):
Nice, would be even better if you included a screenshot of the command being used.

PR #285, opened 2025-01-23
Add attach aws session command

calebstewart:


calebstewart:
Didn't mean to open this here; apologies.

PR #286, opened 2025-01-31
feat: allow spacectl usage in hooks

sephriot:
Confirmed it works on preprod using custom worker:
Hooks:
- `echo "Testing whether machine api token works" > xyz`
- `spacectl stack environment mount --id tofuone file_from_ansible_hook xyz`
image


Clickup-user:
Task linked: [CU-8697guxx8 API token in a stack is not permissive enough for spacectl to run in a hook](https://app.clickup.com/t/8697guxx8)

PR #292, opened 2025-02-27
feat: use headers for tf version files upload

ilya-hontarau:
Send upload headers with the file upload. It's required for the azure provider, other providers will be noop

### DIFF
--- a/client/structs/string_map.go
+++ b/client/structs/string_map.go
@@ -1,5 +1,7 @@
 package structs
 
+import "net/http"
+
 type StringMap struct {
 	Entries []struct {
 		Key   string `json:"key"`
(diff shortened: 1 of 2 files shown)


eliecharra:
@ilya-hontarau Something came to my mind this morning while working on the compatibility matrix. 

Do you think we can do something that will query for the headers only and only if they are available in the GraphQL schema?
That way, new version of spacectl will still be compatible with old installations?
I don't know GraphQL very well, but can we use the spread (`... on sha256SumsUploadHeaders`) operator for that?
Or maybe fetch the schema and switch case on it.

We'll still have to use the last CLI with newer azure installations, but that would reduce by a lot the complexity of maintaining a compatibility matrix.

Do you know what I mean?

ilya-hontarau:
> @ilya-hontarau Something came to my mind this morning while working on the compatibility matrix.
> 
> Do you think we can do something that will query for the headers only and only if they are available in the GraphQL schema? That way, new version of spacectl will still be compatible with old installations? I don't know GraphQL very well, but can we use the spread (`... on sha256SumsUploadHeaders`) operator for that? Or maybe fetch the schema and switch case on it.
> 
> We'll still have to use the last CLI with newer azure installations, but that would reduce by a lot the complexity of maintaining a compatibility matrix.
> 
> Do you know what I mean?

we discussed it offline and as a result, I will create a proposal to the dev team

Issue #289, opened 2025-02-01
Stack local preview is slow in large working trees.


armooo:
The ignore options to local preview still stat and process all files in the working tree. In large monorepo or projects with many ignored files, node_module being one common example, just processing the `matchFn` from `GetIgnoreMatcherFn` takes most of the time. It looks like the matcher support was added in a fork of [archiver](https://github.com/spacelift-io/archiver) here.

One option I see is to change it to return `SkipDir  in the `WalkFunc` if we are skipping a directory. I have tested this out for my example and with #288 the time to create the archive has gone from minutes to milliseconds. If this seems like a good solution to you I can send some PRs.

PR #294, opened 2025-02-28
Do not recurse into ignored directories.

armooo:
When generating the tar.gz for local previews can skip recursively processing them. This potentially saves a lot of processing time by removing the stat calls to the files contained in them. This tends to be the case for node_modules and other commonly ignored paths.

On my internal git repo running on macos this results in a -82% change to the time to generate the tar.tz.
```
$ time spacectl stack local-preview --id test  --no-upload
151.33 secs

$ time ~/src/spacectl/spacectl stack local-preview --id test  --no-upload
26.32 secs
```

I also checked if the same files were included in both tar.gz files.
```
$ tar -tf ~/old.tar.gz | sed 's|[^/]*||' | shasum
82510ad099d3eb9e3c1bba906b8b8dff00c802aa  -

$ tar -tf ~/new.tar.gz | sed 's|[^/]*||' | shasum
82510ad099d3eb9e3c1bba906b8b8dff00c802aa  -
```

Fixes: #289

### DIFF
--- a/internal/cmd/stack/local_preview.go
+++ b/internal/cmd/stack/local_preview.go
@@ -7,7 +7,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/mholt/archiver/v3"
 	"github.com/shurcooL/graphql"
 	"github.com/spacelift-io/spacectl/client/structs"
 	"github.com/spacelift-io/spacectl/internal"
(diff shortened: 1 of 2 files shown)


tomasmik (approved):
I've done some testing and this seem to be working the same as before.

PR #297, opened 2025-03-13
Improve the modules listing command

tomasmik:
image

### DIFF
--- a/internal/cmd/module/list.go
+++ b/internal/cmd/module/list.go
@@ -1,19 +1,17 @@
 package module
 
 import (
+	"context"
 	"fmt"
 	"slices"
 
 	"github.com/pkg/errors"
 	"github.com/shurcooL/graphql"
 	"github.com/spacelift-io/spacectl/client/structs"
+	"github.com/spacelift-io/spacectl/internal"
 	"github.com/spacelift-io/spacectl/internal/cmd"
 	"github.com/spacelift-io/spacectl/internal/cmd/authenticated"
 	"github.com/urfave/cli/v2"
-)
-
-const (
-	maxSearchModulesPageSize = 50
 )
 
 func listModules() cli.ActionFunc {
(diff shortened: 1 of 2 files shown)


internal/cmd/module/list.go
```diff
@@ -1,112 +1,93 @@
...
-			continue
-		}
+	var fullTextSearch *graphql.String
+	if search != nil {
+		fullTextSearch = graphql.NewString(graphql.String(*search))
+	}
 
-		break
+	modules, err := searchModules(ctx.Context, structs.SearchInput{
+		First:          first,
```
frncmx, line 57: This works differently now than the table format, doesn’t it? The limit is not set by default, so it might take the default limit (10) from the API. Wouldn’t that be confusing? A warning would not be printed. I would expect the JSON format to fetch all, whereas with the table output, I would be fine with a limit—provided a message is printed below the table informing me about it.

internal/cmd/module/list.go
```diff
@@ -1,112 +1,93 @@
...
+		}
+
 		switch outputFormat {
 		case cmd.OutputFormatTable:
-			m, err := getModules(cliCtx, 20)
-			if err != nil {
-				return err
-			}
-			return listModulesTable(m)
+			return listModulesTable(cliCtx, search, limit)
```
frncmx, line 36: Should we set a default limit on the table output, such as 10 items, and display the limit below the table?

internal/cmd/module/list.go
```diff
@@ -126,13 +107,51 @@ func listModulesTable(modules []module) error {
...
+		}
+
+		out = append(out, result.Modules...)
+
+		if result.PageInfo.HasNextPage && (fetchAll || limit > len(out)) {
+			pageInput.After = graphql.NewString(graphql.String(result.PageInfo.EndCursor))
+		} else {
+			break
+		}
+	}
```
frncmx, line 147: I wondered if this could be simplified a bit, not sure if it is better, but I thought of something like:

```
	const maxPageSize = 50

	var limit int
	if input.First != nil {
		limit = int(*input.First)
	} else {
		limit = math.MaxInt
	}

	var out []module
	hasNextPage := true
	var after *graphql.String
	for hasNextPage && limit > len(out) {
		pageInput := structs.SearchInput{
			First: graphql.NewInt(
				//nolint: gosec
				graphql.Int(min(limit-len(out),maxPageSize)),
			),
			After:          after,
			FullTextSearch: input.FullTextSearch,
		}

		result, err := searchModules(ctx, pageInput)
		if err != nil {
			return nil, err
		}

		out = append(out, result.Modules...)

		if result.PageInfo.HasNextPage {
			pageInput.After = graphql.NewString(graphql.String(result.PageInfo.EndCursor))
		} else {
			hasNextPage = false
		}
	}

	return out, nil
```
Wdyt? I don't have strong feeling about, but the original code I found a bit hard to understand initially.

frncmx (commented):
I left some comments. My concern is just about the implicit limit on the JSON output.

internal/cmd/module/list.go
tomasmik: It will now default to 20 as it did before.

internal/cmd/module/list.go
tomasmik: > I would expect the JSON format to fetch all, whereas with the table output

It does. We have a clause for that:

> 	fetchAll := limit == 0

internal/cmd/module/list.go
tomasmik: I dont think this code works as you'd expect it to.

internal/cmd/module/list.go
frncmx: Note that you call here `searchModules` and not `searchAllModules`.

internal/cmd/module/list.go
frncmx: Could you also print the limit at the end of table output, so that the user knows they are limited? I think the original confusion came from the fact that we started to limit, but we did not tell. Otherwise, not having the limit is better.

tomasmik: image

frncmx (approved):
thx, lgtm

PR #303, opened 2025-04-09
Add versioning support to spacectl commands

adamconnelly:
I've added a versioning system to spacectl so that we can maintain backwards compatibility with older Self-Hosted instances while still allowing us to add new functionality. The way it works is by creating a wrapper around the CLI commands which understands the concept of the minimum Spacelift version required. At startup we ask for the version of the Spacelift instance and then resolve the available commands.

This means that customers won't need to worry about what version of Spacectl they run against their Spacelift instance in future - it should just work.

The way it fits into the Self-Hosted release cycle is like this:

1. A new command version is added that relies on GraphQL API functionality not released to any Self-Hosted version yet.
1.1. The new version is marked with `SupportedVersionLatest` so that it is only available to SaaS.
2. Self-Hosted version 1.2.3 is released.
2.1. At this point we update `SupportedVersionLatest` to `SupportedVersion("1.2.3")`, allowing it to be used with the latest version of Self-Hosted.
2.2. Older Self-Hosted versions continue to use the previous version of the command.

There's more details in the [contributing guide](https://github.com/spacelift-io/spacectl/pull/303/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055).

I'm happy to split this PR into smaller chunks - I just want to check the concept makes sense first.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,177 @@
+# spacectl Contributing Guide
+
+In this guide you will get an overview of the contribution workflow from opening an issue, creating a pull request, to having it reviewed, merged and deployed.
+
+We welcome all contributions, no matter the size or complexity. Every contribution helps and is appreciated.
+
+The following is a set of guidelines for contributing to spacectl. These are mostly guidelines, not rules. Use your best judgment, and feel free to propose changes to this document in a pull request.
+
+## Reporting an Issue
+
+If you spot a problem with spacectl [search if an issue already exists](https://github.com/spacelift-io/spacectl/issues). If a related issue doesn't exist, please open a new issue using [the issue form](https://github.com/spacelift-io/user-documentation/issues/new).
+
+## Making Changes
+
+spacectl is written in Go, and uses the [Spacelift GraphQL API](https://docs.spacelift.io/integrations/api) to interact with Spacelift. To make changes to spacectl you will need at least the following:
+
+- A working copy of [Go](https://go.dev/). Check [go.mod](./go.mod) to see what is the current version used by this project.
+- A Spacelift account you can use for testing. Please go to <https://spacelift.io/free-trial> to create a free account if you don't already have one.
+
+### Command versioning
+
+spacectl supports working with both SaaS and Self-Hosted instances of Spacelift. Because of this we need to have a mechanism in place for ensuring API compatibility with different versions. To support this, each command can have multiple different implementations, which work with different versions of Spacelift.
+
+#### Defining a command
+
+Commands are defined using the `cmd.Command` struct like in the following example:
+
+```go
+{
+	Category: "Run local preview",
+	Name:     "local-preview",
+	Usage:    "Start a preview (proposed run) based on the current project. Respects .gitignore and .terraformignore.",
+	Versions: []cmd.VersionedCommand{
+		{
+			EarliestVersion: cmd.SupportedVersionAll,
+			Command: &cli.Command{
+				Flags: []cli.Flag{
+					flagStackID,
+					// ... other flags
+				},
+				Action:    localPreview(false),
+				Before:    authenticated.Ensure,
+				ArgsUsage: cmd.EmptyArgsUsage,
+			},
+		},
+		{
+			EarliestVersion: cmd.SupportedVersion("2.5.0"),
+			Command: &cli.Command{
+				Flags: []cli.Flag{
+					flagStackID,
+					// ... other flags
+				},
+				Action:    localPreview(true),
+				Before:    authenticated.Ensure,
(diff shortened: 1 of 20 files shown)


adamconnelly:
The reason for the lint failures is this: https://github.com/spacelift-io/spacectl/pull/304.

ilya-hontarau:
Thank you, interesting idea.
Just to double-check, this implementation doesn't cover the problem when we want to extend the existing command with a new query field?

adamconnelly:
> Thank you, interesting idea. Just to double-check, this implementation doesn't cover the problem when we want to extend the existing command with a new query field?

It should solve it @ilya-hontarau. In that case you just create a new command version. Take a look at the module local-preview command for an example: https://github.com/spacelift-io/spacectl/pull/303/files#diff-7bc44780445cacf5025cfd6eb419e3ec7ecf28e822a9491b353617db2687bc6cR64

It only asks for the upload headers if the SH version supports them.

main.go
```diff
@@ -23,31 +28,90 @@ import (
...
+
 	app := &cli.App{
 		Name:                 "spacectl",
 		Version:              version,
 		Compiled:             compileTime,
 		Usage:                "Programmatic access to Spacelift GraphQL API.",
 		EnableBashCompletion: true,
-		Commands: []*cli.Command{
-			module.Command(),
+		Commands: append([]*cli.Command{
```
tomasmik, line 100: Really good that you thought about these from the start.

main.go
```diff
@@ -23,31 +28,90 @@ import (
...
+func getSpaceliftInstanceVersion() cmd.SpaceliftInstanceVersion {
+	// Default fallback when we can't determine the version
+	instanceVersion := cmd.SpaceliftInstanceVersion{
+		InstanceType: cmd.SpaceliftInstanceTypeUnknown,
+	}
+
+	ctx, httpClient := session.Defaults()
+
+	// Create a new session - this may fail if the user doesn't have valid credentials
+	sess, err := session.New(ctx, httpClient)
```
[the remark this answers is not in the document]
adamconnelly: We could do - I just wanted to avoid outputting an error if it was totally valid usage by customers (e.g. you've just installed spacectl and haven't setup credentials, but you want to run the `--help` command).

In this case what will happen is it'll output a warning here: https://github.com/spacelift-io/spacectl/pull/303/files/341122b46ce3ee2746ae181d2b9e1f6aebee07cb#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R90.

Maybe what we should do is return sentinel errors from `session.New()` that we can handle properly here so we can distinguish between the session just not being configured vs some other random error?

main.go
adamconnelly: Yeah - while I was working on this I was worried we'd hit a chicken and egg situation. But I think it ended up working out fine, and we just have some core commands that are always available.

main.go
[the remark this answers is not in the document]
tomasmik: That sounds reasonable, but also maybe a bit out of scope 🤷🏻

CONTRIBUTING.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Reporting an Issue
+
+If you spot a problem with spacectl [search if an issue already exists](https://github.com/spacelift-io/spacectl/issues). If a related issue doesn't exist, please open a new issue using [the issue form](https://github.com/spacelift-io/user-documentation/issues/new).
+
+## Making Changes
+
+spacectl is written in Go, and uses the [Spacelift GraphQL API](https://docs.spacelift.io/integrations/api) to interact with Spacelift. To make changes to spacectl you will need at least the following:
+
+- A working copy of [Go 1.22](https://go.dev/).
```
eliecharra, line 17: We could just as to refer to the go.mod instead of pinning the version here.

```suggestion
- A working copy of [Go](https://go.dev/). Check `go.mod` to see what is the current version used by this project.
```

Just a nit, to save some maintenance effort.

main.go
```diff
@@ -24,31 +29,90 @@ import (
 var version = "dev"
 var date = "2006-01-02T15:04:05Z"
 
+type debugInfoQuery struct {
+	DebugInfo struct {
+		SelfHostedVersion string `graphql:"selfHostedVersion"`
+	} `graphql:"debugInfo"`
+}
+
+func getSpaceliftInstanceVersion() cmd.SpaceliftInstanceVersion {
```
ilya-hontarau, line 38: nit: I think some comments can be removed

ilya-hontarau (approved):
Thanks, looks great, I had actually in mind more complex solutions, it's much simpler

PR #309, opened 2025-04-19
fix: update `provider create-version` to use versioned commands

adamconnelly:
The command was already backwards-compatible because we were parsing the GraphQL schema to check if the mutation was available, but I'm changing it to use the command versioning approach instead so we're not using multiple approaches to backwards compatibility.

The new `terraformProviderVersionRegisterPlatformV2` command was introduced in Self-Hosted version 2.3.0, so updated spacectl to only use that version of the command for 2.3.0 or later.


adamconnelly:
I'm converting this to a draft - I'm going to tweak the approach slightly.

adamconnelly:
Ok - that's it updated now to handle both the `terraformProviderVersionRegisterPlatformV2` mutation as well as the new upload header fields being returned from GraphQL responses.

ilya-hontarau (approved):
Thank you

tomasmik (approved):

PR #310, opened 2025-04-23
Implement MCP server command

Axot017:
Initial version of Spacelift MCP server

This PR adds the first implementation of the Spacelift MCP server with several tools:
- listing stacks
- listing stack runs
- showing run changes
- showing run logs
- triggering run
- discarding run
- approving run
- showing resources
- local preview

It's still a draft and needs some cleanup - the tool inputs and outputs could use better descriptions, and there might be some code organization improvements to make. But I think there's enough here to get
feedback or suggestions on the approach.

Let me know what you think!

Demo with claude code:

https://github.com/user-attachments/assets/5335098b-6a78-48b6-9253-0f495b359d6b


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/spacelift-io/spacectl
 
-go 1.22
+go 1.23.4
```
Axot017, line 3: `github.com/mark3labs/mcp-go` requires go 1.23. Are we ok with this change?

tomasmik: Sure.

tomasmik (commented):
If I understand the correct usage case for this, wouldn't the local preview commands be huge as well? Make changes locally and have the AI be able to test them without committing code.

go.mod
mbialon: Should we bump the major version on release?

README.md
```diff
@@ -269,6 +269,12 @@ By default the login process is interactive, however, if that does not fit your
 
 You can switch between account profiles by using `spacectl profile select ${MY_ALIAS}`. What this does behind the scenes is point `${HOME}/.spacelift/current` to the new location. You can also delete stored credetials for a given profile by using the `spacectl profile logout ${MY_ALIAS}` command.
 
+## MCP Server
```
mbialon, line 272: Could you make the instructions more detailed? Or link to a docs page with specific instructions?

go.mod
Axot017: It is already bumped on main

README.md
Axot017: I added some details

Axot017:
It should be ready to review and merge now. @tomasmik I added also local preview tool as you suggested

mbialon (approved):
LGTM

README.md
```diff
@@ -269,6 +269,63 @@ By default the login process is interactive, however, if that does not fit your
...
+```
+
+Or if you prefer using Docker:
+
+```json
+{
+  "mcpServers": {
+    "spacelift": {
+      "command": "docker",
+      "args": ["run", "-i", "--rm", "ghcr.io/spacelift-io/spacectl", "mcp", "server"]
```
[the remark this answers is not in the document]
Axot017: Good point, we should clarify that. I added an example to README. It can be done using -e flag to pass env variables

tomasmik (approved):
let's do a release after

PR #314, opened 2025-05-06
chore(golangcilint): delete golangcilint v1 backup config file

peterdeme:
The migration tool creates a backup file. Which we don't need.


tomasmik (approved):
one step ahead of me this time. nice.

Issue #316, opened 2025-05-06
Add preview runs to stack run list


zbstof:
I'm creating new run
```
spacectl stack preview --id my-stack --sha "${{ github.sha }}" --tail
```
but I don't see its details being listed when i do
```
spacectl stack run list --id my-stack
```
I assume stack run list doesn't return preview runs currently. Can this feature be added?

zbstof:
Or any other workaround to get ID of the preview run from the command line would be nice

Axot017:
Hey @zbstof, we just released [v1.13.0](https://github.com/spacelift-io/spacectl/releases/tag/v1.13.0) which allows you to filter for preview runs using the --preview-runs flag like this:
```
spacectl stack run list --id my-stack --preview-runs
```
This will show only preview runs for the specified stack. Let me know if you have any questions or encounter any issues.

PR #317, opened 2025-05-08
feat: Add --preview-runs flag to runs list command

Axot017:
Fixes: https://github.com/spacelift-io/spacectl/issues/316 

image

### DIFF
--- a/internal/cmd/stack/run_list.go
+++ b/internal/cmd/stack/run_list.go
@@ -29,9 +29,19 @@ func runList(cliCtx *cli.Context) error {
 
 	switch outputFormat {
 	case cmd.OutputFormatTable:
-		return listRunsTable(cliCtx.Context, stackID, maxResults, showPreview)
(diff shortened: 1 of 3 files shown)


internal/cmd/stack/run_list.go
```diff
@@ -65,37 +67,86 @@ type runsJSONQuery struct {
...
 	var before *string
 
 	for len(results) < maxResults {
-		var query struct {
-			Stack *struct {
-				Runs []runsJSONQuery `graphql:"runs(before: $before)"`
-			} `graphql:"stack(id: $stackId)"`
+		var err error
+		var runs []T
+		if preview {
```
mbialon, line 122: Could we pass a "fetcher" function here rather than a bool? We could decide which function to pass in the `runList` function.

It would make the `fetchRuns` and `listRuns(Table,JSON)` functions easier to test.

Axot017: I like that. I'll change it in a moment

mbialon (approved):
LGTM

Issue #95, opened 2022-09-19
Stack: local-preview unauthorized suggests re-logging in even with a valid token


lexton:
The situation plays out as so:
* User has a valid login session
* User doesn't have write access to a stack

User is presented with an error + suggested fix that doesn't resolve the issue:
```
unauthorized: you can re-login using `spacectl profile login`
```

It would be better to say - you don't have access to this resource, or even this resource doesn't exist rather than suggesting to log in again.

cube2222:
Good point. Maybe when we're about to display such an error we should make a graphql call akin to "whoami" to see if we have no working session, or whether the session just doesn't have a high enough access level.

Issue #65, opened 2022-05-24
Unauthorized errors are misleading


dmikalova:
My team recently started using SpaceLift and I was trying out the local preview feature. However I get an unauthorized error:

```
spacectl stack preview --id x
2022/05/24 14:57:04 unauthorized: you can re-login using `spacectl profile login`
```

The error makes it seem like I am not logged in, but the following works:

```
spacectl profile current
spacectl stack show --id x
```

After talking with SpaceLift support it seems like I mistakenly don't have permissions to run a stack which I can address with my SpaceLift admin.

It would be helpful if the spacectl errors could differentiate between not being logged in and not having permissions to run a stack. Thanks!

PR #329, opened 2025-05-29
fix the error printed when user has no access

tomasmik:
## Description

This was confusing to people because they though they need to login, but actually they got no access.
The error is not saying that this is the case 100%, just so that we dont lie by accident. There isn't an easy way to determine this tbh.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I have added/updated tests as needed
- [x] All existing tests pass

## Screenshots (if applicable)

Add screenshots to show visual changes.

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code where necessary
- [x] My changes generate no new warnings

## Related Issues

Closes https://github.com/spacelift-io/spacectl/issues/65
Closes https://github.com/spacelift-io/spacectl/issues/95

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -30,12 +30,7 @@ func (c *client) Mutate(ctx context.Context, mutation interface{}, variables map
 	}
 
 	err = apiClient.Mutate(ctx, mutation, variables, opts...)
-
-	if err != nil && err.Error() == "unauthorized" {
-		return fmt.Errorf("unauthorized: you can re-login using `spacectl profile login`")
-	}
-
-	return err
+	return c.determineClientError(ctx, apiClient, opts, err)
 }
 
 func (c *client) Query(ctx context.Context, query interface{}, variables map[string]interface{}, opts ...graphql.RequestOption) error {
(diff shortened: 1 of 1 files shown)


client/client.go
```diff
@@ -45,12 +40,29 @@ func (c *client) Query(ctx context.Context, query interface{}, variables map[str
...
+	if !strings.Contains(err.Error(), "unauthorized") {
+		return err
+	}
+
+	var query struct {
+		Viewer *struct {
+			ID string `graphql:"id" json:"id"`
+		}
+	}
+	queryErr := client.Query(ctx, &query, map[string]interface{}{}, opts...)
```
tomasmik, line 60: I understand that this isn't very DRY as we're almost 100% repeating ourselves here, however I don't want to deal with crazy import cycles or nonsensical packages. This is good.

mbialon (approved):

Issue #357, opened 2025-10-30
spacectl login w/ env variables set leads to login issues when utilizing browser login


Anzor-Ishak:
possibly related to:
https://github.com/spacelift-io/spacectl/issues/270

User reported:

Behavior is not consistent utilizing spacectl login.
Sometimes able to login, sometimes not able to. 

**The login on the web says the authentication flow has successfully completed.
But cannot do anything**

Ex:
"Waiting for login...
Done!"

"failed to query user information: unauthorized"
"You are not logged in, could not find stacks"

**Workaround**
Remove environment variables (had valid API key) and login works fine.

PR #358, opened 2025-10-31
Print a warning if you login but have env set

tomasmik:
## Description

This will print a warning and show you which variables are set after you login.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Related Issues

Closes https://github.com/spacelift-io/spacectl/issues/357

### DIFF
--- a/internal/cmd/profile/login_command.go
+++ b/internal/cmd/profile/login_command.go
@@ -38,6 +38,8 @@ func loginCommand() *cli.Command {
 }
 
 func loginAction(ctx context.Context, cliCmd *cli.Command) error {
+	defer printEnvWarning()
+
 	var storedCredentials session.StoredCredentials
 
 	// Let's try to re-authenticate user.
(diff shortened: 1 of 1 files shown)


internal/cmd/profile/login_command.go
```diff
@@ -211,3 +213,39 @@ func persistAccessCredentials(creds *session.StoredCredentials) error {
...
+func checkEnvironmentVariables() []string {
+	var envVars []string
+
+	checkVars := []string{
+		session.EnvSpaceliftAPIToken,
+		session.EnvSpaceliftAPIKeyEndpoint,
+		session.EnvSpaceliftAPIEndpoint,
+		session.EnvSpaceliftAPIGitHubToken,
+		session.EnvSpaceliftAPIKeyID,
+		session.EnvSpaceliftAPIKeySecret,
```
eliecharra: Aren't we going to print secret credentials to stdout with that? 🤔

tomasmik: That'd be funny.

We just print the name of the var:
```
WARNING: The following Spacelift environment variables are set:
   - SPACELIFT_API_ENDPOINT

Environment variables take precedence over profile credentials.
If these credentials are expired or invalid, commands will fail even after successful login.
```

Issue #362, opened 2025-12-19
[FEATURE] Only show stacks with local preview enabled when run with no stack ID


med-94:
## Feature Description

Currently, when running `spacectl stack local-preview` with no arguments the CLI will list all stacks associated with that project, even if they don't have local preview enabled. I would like it so that it only showed stacks where local preview was enabled.

## Problem Statement

We have some repos with a very large amount of stacks, but local preview is only enabled on a subset of them. It would be more ergonomic if when running a local preview only stacks where it was possible to trigger showed up.

## Proposed Solution

Looking at the code, I think this could be achieved by passing an additional query predicate through to the graphql query that gets made to search for stacks.

PR #363, opened 2025-12-19
Only show stacks with local preview enabled

med-94:
## Description

Attempt at an implementation for https://github.com/spacelift-io/spacectl/issues/362

This doesn't work currently as the `localPreviewEnabled` field doesn't seem to be valid for searching stacks:
```
2025/12/19 11:56:36 failed search for stacks: could not parse search query: couldn't parse 2nd predicate: unknown field: localPreviewEnabled
```

Submitting this in case it's a simple change that can be made (wrong field name or similar), but I couldn't find docs on the list of accepted predicates when searching stacks.

I'm also not a massive fan of the additional argument to `getStack` but couldn't see an easy way to avoid it. I've made it accept a slice of predicates to try and be as generic as possible in case there are other places where we'd want to pre-filter the stacks when showing the selector

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changes Made

- Adds new `extraConditions` argument to `getStack`, which accepts a slice of `QueryPredicate` which are then appended to the graphql call made when searching for stacks.
- When running local-preview adds a predicate that (attempts to!) filters for stacks with local preview enabled

## Testing

- [ ] I have added/updated tests as needed
- [x] All existing tests pass

## Screenshots (if applicable)

Add screenshots to show visual changes.

## Checklist

- [ ] My code follows the project's style guidelines
- [ ] I have performed a self-review of my code
- [ ] I have commented my code where necessary
- [ ] My changes generate no new warnings

## Related Issues

Closes #(issue number)
Fixes #(issue number)
Related to #(issue number)

### DIFF
--- a/internal/cmd/stack/delete.go
+++ b/internal/cmd/stack/delete.go
@@ -27,7 +27,7 @@ var flagSkipConfirmation = &cli.BoolFlag{
 
 func deleteStack() cli.ActionFunc {
 	return func(ctx context.Context, cliCmd *cli.Command) error {
-		stackID, err := getStackID(ctx, cliCmd)
+		stackID, err := getStackID(ctx, cliCmd, nil)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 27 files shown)


tomasmik:
Hi thanks for the submission and you are right, there is no way to filter according to this parameter. I'll talk with the team on what we can do about this.

PR #388, opened 2026-03-06
fix: use caFile variable instead of hardcoded "cacert" in configureTLS

bernardm-k:
## Summary

`configureTLS()` reads `SPACELIFT_API_TLS_CA` into `caFile` but passes the hardcoded string `"cacert"` to `os.ReadFile`, making the CA bundle feature non-functional. This also adds `SSL_CERT_FILE` as a fallback.

### `internal/cmd/authenticated/client.go`

**Problem:** Line 84 calls `os.ReadFile("cacert")` instead of `os.ReadFile(caFile)`. Setting `SPACELIFT_API_TLS_CA=/etc/ssl/cert.pem` tries to read a literal file named `cacert` in the current directory.

**Solution:**
- Fix `os.ReadFile` to use the `caFile` variable
- Add `SSL_CERT_FILE` env var as a fallback when `SPACELIFT_API_TLS_CA` is not set
- `SPACELIFT_API_TLS_CA` takes precedence when both are set

**Impact:** Unblocks spacectl usage in sandboxed macOS environments (e.g., restricted CI runners, containerized dev environments) where Security.framework cannot access the Keychain, causing `x509: OSStatus -26276`. Setting `tls.Config.RootCAs` to a non-nil pool forces Go's pure-Go certificate verifier, bypassing the Security.framework dependency entirely.

### `internal/cmd/authenticated/client_test.go` (new)

Adds 6 test cases for `configureTLS`:
- No env vars set → nil RootCAs (default system behavior)
- `SPACELIFT_API_TLS_CA` set → RootCAs populated
- `SSL_CERT_FILE` fallback → RootCAs populated
- Both set → `SPACELIFT_API_TLS_CA` takes precedence
- Invalid CA path → error returned
- Invalid certificate content → `errEnvSpaceliftAPIClientCAParse`

## Net Impact

- Fixes a bug where `SPACELIFT_API_TLS_CA` has never worked
- Adds standard `SSL_CERT_FILE` support used by other Go tooling
- No breaking changes — existing behavior without env vars is unchanged

## Test plan

- [ ] `go test ./internal/cmd/authenticated/...` passes
- [ ] Set `SPACELIFT_API_TLS_CA=/etc/ssl/cert.pem` and verify `spacectl profile list` works in a sandboxed environment
- [ ] Verify no env vars set still uses system defaults

### DIFF
--- a/internal/cmd/authenticated/client.go
+++ b/internal/cmd/authenticated/client.go
@@ -80,8 +80,19 @@ func configureTLS(httpClient *http.Client) error {
 		MinVersion: tls.VersionTLS12,
 	}
 
-	if caFile, ok := os.LookupEnv(EnvSpaceliftAPIClientCA); ok && caFile != "" {
-		caCert, err := os.ReadFile("cacert")
+	// Check SPACELIFT_API_TLS_CA first, then fall back to SSL_CERT_FILE.
+	// Setting RootCAs to a non-nil pool forces Go's pure-Go certificate
+	// verifier, bypassing Security.framework (which fails in sandboxed
+	// environments where macOS Keychain is inaccessible).
+	caFile := ""
+	if f, ok := os.LookupEnv(EnvSpaceliftAPIClientCA); ok && f != "" {
+		caFile = f
+	} else if f, ok := os.LookupEnv("SSL_CERT_FILE"); ok && f != "" {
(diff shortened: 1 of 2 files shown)


bernardm-k:
I believe it is fixed now.

PR #407, opened 2026-03-28
Allow mutations in the `api` command

alexjurkiewicz:
This allows `spacectl api` to execute arbitrary mutations, but only when `--mutation` is passed explicitly.

It supports both shorthand and full mutation syntax:

```bash
spacectl api --mutation 'stackDelete(id: "x") { id }'
spacectl api --mutation 'mutation { stackDelete(id: "x") { id } }'
```

### DIFF
--- a/internal/cmd/api/README.md
+++ b/internal/cmd/api/README.md
@@ -1,8 +1,8 @@
 # `spacectl api`
 
-`spacectl api` lets you run ad-hoc read-only GraphQL queries against the Spacelift API using your existing authentication.
+`spacectl api` lets you run ad-hoc GraphQL operations against the Spacelift API using your existing authentication.
 
-Mutations are not supported. For write operations, use the dedicated `spacectl` subcommands or the [Spacelift Terraform Provider](https://github.com/spacelift-io/terraform-provider-spacelift).
+Queries are allowed by default. Mutations require the explicit `--mutation` flag as a safety guard. Subscriptions are not supported.
 
 ## Usage
 
@@ -25,12 +25,20 @@ With variables:
 spacectl api --variables '{"id":"my-stack"}' 'query($id: ID!) { stack(id: $id) { id name } }'
 ```
 
+Mutations require `--mutation`. Bare mutation field selections are wrapped in `mutation { ... }` automatically:
+
+```bash
+spacectl api --mutation 'stackDelete(id: "my-stack") { id }'
+spacectl api --mutation 'mutation DeleteStack($id: ID!) { stackDelete(id: $id) { id } }' --variables '{"id":"my-stack"}'
+```
+
 From a file or stdin:
 
 ```bash
 spacectl api < query.graphql
 spacectl api --variables '{"id":"my-stack"}' < query.graphql
 cat query.graphql | spacectl api
+spacectl api --mutation < mutation.graphql
 ```
 
 ## Output
--- a/internal/cmd/api/api.go
+++ b/internal/cmd/api/api.go
@@ -23,6 +23,10 @@ var (
 		Name:  "variables",
 		Usage: "JSON object with GraphQL variables",
 	}
+	flagMutation = &cli.BoolFlag{
+		Name:  "mutation",
+		Usage: "Allow and execute a GraphQL mutation",
+	}
 	flagRaw = &cli.BoolFlag{
 		Name:  "raw",
 		Usage: "Output response body without pretty-printing",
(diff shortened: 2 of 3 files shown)


alexjurkiewicz:
This is the other major part of my too-complex proposal in #391. I hope it can be accepted. It's been extremely useful locally, Claude, Copilot and Codex have all been much better at driving `spacectl api '...'` than they have `curl ...`.

One major reason is they don't need to deal with authentication. It's impossible for AIs to correctly authenticate with the Spacelift API without documentation, and even once they have a bearer token, often screw it up.

A related pain point: most AI harnesses require explicit user authorisation for any command with `$(cat token.txt)` command expansions embedded. Which means every GraphQL call using `curl -H "Authorization: Bearer $(cat spacelift-token.txt)" ...` requires manual approval 🫠

eliecharra:
Hey @alexjurkiewicz 👋🏻  Sorry for the delay. 
The team owning spacectl has been quite busy lately but we should have come back to you earlier 🙇🏻

We're thinking about the best way to make usage of agents easier with spacectl, and we'll come back to you in a few days probably. We do not mind improving this experience, we're just not sure if this implementation is the best way we want to enable that.

Do you think that configuring spacectl as a local MCP server could be a better story for helping improve agent usage?
We have this [section](https://docs.spacelift.io/integrations/api-development-with-mcp#spacelift-api-integration-via-spacectl-mcp-server) of the docs about how to do that. So maybe we can just improve what does the MCP server exposes, and that's it?

alexjurkiewicz:
Long term I'm sure there is a better solution. But I want my agent to interact with Spacelift features _right now_.

Your better solutions might not arrive for months.

`spacectl api` is a useful escape hatch for performing arbitrary queries against the API. In the same way that `gh api` is useful.

If you had the engineering resources of AWS, you could build a CLI that supports every endpoint automatically. But not even Github can do that, so they add the `gh api` endpoint.

Here are some recent tasks I accomplished with `spacectl api` queries:
* List AWS integrations to understand where duplicates exist in our infra
* List stack dependencies to help graph our dependencies in MermaidJS

Here are some recent tasks I've accomplished with `spacectl api --mutation` queries:
* Migrate stacks and their many attached policies/contexts to another space, one-by-one
* Create new blueprints

You might eventually add support for all these use cases. But I don't want to wait months

alexjurkiewicz:
I don't want to belabour the point, but you can't keep up with spacectl features. This year `spacectl` has shipped support for these new API endpoints:

* `spacectl stack promote` -> `runPromote` mutation
* `spacectl provider version docs upload` -> `terraformProviderVersionUploadDocs` mutation
* `spacectl stack state pull` -> `stateDownloadUrl` mutation
* `spacectl stack task list` -> `stack { tasks` query

(According to Claude's searching). You won't add CLI support for the featuers I used above any time soon. Enhancing `spacectl api` is how to cover the long-tail of feature requests in a timely fashion.

eliecharra:
I got your point @alexjurkiewicz. I think you got me slightly wrong.

What I was suggesting here is that instead of creating a new CLI  subcommand, we could add such a feature to the existing MCP server directly. So your agent could directly call the spacectl local MCP server to run its queries and mutations. Some kind of new `RunRawGraphQLQuery` capability to the existing MCP server instead of a CLI subcommand, but with the same logic. It's more "AI native" if you see what I mean.

My concern was more on the implementation detail. I'm not challenging the overall idea to have a passthrough auth proxy that could follow up quickly with backend API changes.

alexjurkiewicz:
ok, I see 😊 

I'm not that fussed about implementing this as an MCP tool compared to a CLI command.

I think that in general the tide is turning away from MCP tools, and towards CLI tools with skills to document their usage. Implementing this as a CLI has several benefits:
* The feature works for any consumer, not just AI agents. For example, people might want to make arbitrary API calls in CI, in shell scripts, or as provisioning hooks in Terraform
* MCP servers aren't shared between different coding agents. If I have a claude code session in my terminal and copilot in vscode I need two `spacectl` mcp servers running

Personally, I don't feel strongly. But MCP servers seem to be less cool lately.

tomasmik (changes requested):
Hi, thanks for the submission. We've reviewed the changes and have decided to not accept this code, however we can accept the concept of allowing mutations or anything to go through this same command.

Could you open a new, simplified version of this change, where when provided with a flag, we do not filter not do we check the type of the request. Basically there are no safety checks anymore. This way we reduce the code by a lot and enable what you want. Basically lets get rid of `normalizeDocument`, `firstSignificantToken` and instead of attempting to enforce something arbitrary lets allow no safe guards.

tomasmik:
Please open a new PR, thanks.

PR #419, opened 2026-04-17
Allow mutations in the `api` command

alexjurkiewicz:
Follow-up to #407.

This allows `spacectl api` to execute arbitrary GraphQL documents, including mutations.

The provided document is sent unchanged to the API.

```bash
spacectl api '{ stacks { id name } }'
spacectl api 'mutation { stackDelete(id: "x") { id } }'
```

### DIFF
--- a/internal/cmd/api/README.md
+++ b/internal/cmd/api/README.md
@@ -1,16 +1,16 @@
 # `spacectl api`
 
-`spacectl api` lets you run ad-hoc read-only GraphQL queries against the Spacelift API using your existing authentication.
+`spacectl api` lets you run ad-hoc GraphQL operations against the Spacelift API using your existing authentication.
 
-Mutations are not supported. For write operations, use the dedicated `spacectl` subcommands or the [Spacelift Terraform Provider](https://github.com/spacelift-io/terraform-provider-spacelift).
+The provided document is sent unchanged to the API, so queries and mutations are both supported.
 
 ## Usage
 
-Basic queries (bare field selections are wrapped in `query { ... }` automatically):
+Basic queries:
 
 ```bash
-spacectl api 'stacks { id name state }'
-spacectl api 'workerPools { id name workers { id } }'
+spacectl api '{ stacks { id name state } }'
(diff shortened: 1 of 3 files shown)


alexjurkiewicz:
I wasn't sure about the best argument name. `--literal`, `--mutation`, `--raw-input`, rename `--raw` to `--raw-output`, so many candidates... 🤯

tomasmik:
> I wasn't sure about the best argument name. `--literal`, `--mutation`, `--raw-input`, rename `--raw` to `--raw-output`, so many candidates... 🤯

That's why i didnt want to suggest a name in the original request :( I'll think a bit about possible names, but this one isn't terrible.

tomasmik:
maybe `--any`? @alexjurkiewicz ?

alexjurkiewicz:
I don't love `--any`. I think a lot of these ideas suffer from the same problem, it's not clear they refer to the input query. `--literal` or `--any` -- literal what? Literal formatting? Literal output?

The other alternative is to copy `gh api`. Allow reads and writes by default. Users are aware when they run a mutation because the input says `mutation {`. Then we don't need a flag or the simple `mutation` detection at all!

alexjurkiewicz:
`gh api` is my guiding star for this feature. It's such a great interface, albeit for a REST API. I think we should imitate it here. They don't need a flag for mutations/POSTs.

tomasmik:
`--allow-any`, `--unsafe`? @alexjurkiewicz

alexjurkiewicz:
`-allow-mutation` (no s) is my vote

tomasmik:
Seeing how this is used and that no name really fits, I'm willing to say that no extra flag is needed. Let's just remove the check where we only allow query request and lets just treat this command as a wrapper for the api with auth functionality.

alexjurkiewicz:
ok, the wrapper is as thin as thin can be!

tomasmik (approved):

alexjurkiewicz:
thank you for your guidance over the past month on this feature Tomas. I'm really happy it's merged <3 looking forward to the next release!

PR #436, opened 2026-05-15
Add worktools and more to show command

tomasmik:
## Description

Because I was extending the config, I also extended it to have other vendors too.

### DIFF
--- a/internal/cmd/stack/show.go
+++ b/internal/cmd/stack/show.go
@@ -16,8 +16,13 @@ import (
 )
 
 const (
-	stackConfigVendorPulumi    = "StackConfigVendorPulumi"
-	stackConfigVendorTerraform = "StackConfigVendorTerraform"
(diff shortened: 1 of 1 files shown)


internal/cmd/stack/show.go
```diff
@@ -373,15 +444,47 @@ func (c *showStackCommand) outputModuleVersionUsage(query showStackQuery) error
...
+		return "Terraform"
+	case "OPEN_TOFU", "OPENTOFU":
+		return "OpenTofu"
+	case "KUBERNETES":
+		return "kubectl"
+	case "MANUALLY_PROVISIONED":
+		return "Manually provisioned"
+	case "CUSTOM":
+		return "Custom"
+	case "":
```
mbialon: Nit: this branch is covered by `return tool`

mbialon (approved):

PR #446, opened 2026-06-08
feat: allow overriding config dir via SPACELIFT_CONFIG_DIR

peterdeme:
## What

Adds `SPACELIFT_CONFIG_DIR`, a first-class environment variable that overrides where `spacectl` stores its profiles and configuration. When set, `UserProfileManager` uses it; otherwise it falls back to the existing `${HOME}/.spacelift` default.

```bash
export SPACELIFT_CONFIG_DIR=/workspace/.spacelift
spacectl profile login my-account
```

## Why

The profile/config directory was hardcoded to `${HOME}/.spacelift`. In containers and CI runners `$HOME` is often unset or read-only, and the only workaround was to override `HOME` itself, which is a surprising side effect that also relocates everything else reading it.

`UserProfileManager` is the single entry point every profile-reading command goes through, so honoring the variable there covers the whole CLI without threading a flag through every command.

> [!NOTE]
> Behavior is unchanged when the variable is unset, so this is fully backwards compatible.

## Testing

- Added a test asserting the manager resolves to the custom dir when `SPACELIFT_CONFIG_DIR` is set.
- `go build ./...`, `go vet ./client/session/...`, and `go test ./client/session/...` pass; `gofmt` clean.

Tracking: [CU-869dkyp3r](https://app.clickup.com/t/869dkyp3r)


Clickup-user:
Task linked: [CU-869dkyp3r spacectl: make TLS CA and config-file location configurable for containerized/CI use](https://app.clickup.com/t/869dkyp3r)

haluz (approved):

PR #467, opened 2026-07-24
Add the logs command for modules

tomasmik:
## Description

```
./spacectl module logs --id terraform-default-module-good --run 01KY9RP8WG9V6N1VNHS1V7DX7M

-----------------
QUEUED	Fri Jul 24 12:53:17 EEST 2026
-----------------


-----------------
READY	Fri Jul 24 12:53:17 EEST 2026
-----------------


-----------------
PREPARING	Fri Jul 24 12:53:22 EEST 2026
-----------------

[Ground Control] This is Ground Control to public worker 01KY9QN50JCM4GMSQPSSJGSV9P, commencing countdown
[Ground Control] No initialization policies attached
[Ground Control] Creating metadata...
[Ground Control] Uploading metadata...
[Ground Control] Metadata is GO
[Ground Control] Creating workspace...
[Ground Control] Encrypting workspace...
[Ground Control] Uploading workspace...
[Ground Control] Workspace is GO
```

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/internal/logs/explorer.go
+++ b/internal/logs/explorer.go
@@ -29,14 +29,24 @@ type Explorer struct {
 	backoff time.Duration
 }
 
-// NewExplorer creates a new Explorer with the given options.
-// By default the explorer explores a stack's run and always tails the logs;
-// use WithModule to explore a module's run instead.
-func NewExplorer(id, run string, opts ...Option) *Explorer {
+// NewStackExplorer creates a new Explorer for a run belonging to a stack.
+// By default the explorer always tails the logs.
+func NewStackExplorer(stackID, run string, opts ...Option) *Explorer {
+	return newExplorer(entityStack, stackID, run, opts...)
+}
+
+// NewModuleExplorer creates a new Explorer for a run belonging to a module.
+// By default the explorer always tails the logs.
+func NewModuleExplorer(moduleID, run string, opts ...Option) *Explorer {
+	return newExplorer(entityModule, moduleID, run, opts...)
+}
+
+func newExplorer(entity entity, id, run string, opts ...Option) *Explorer {
 	e := &Explorer{
 		id:      id,
 		run:     run,
 		tail:    true,
+		entity:  entity,
 		backoff: 0,
 	}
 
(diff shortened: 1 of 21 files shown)


internal/logs/explorer.go
```diff
@@ -212,9 +202,18 @@ func (e *Explorer) actionFunc(state structs.RunState) error {
...
 }
+
+// label returns a human-readable name for the entity whose run is being explored.
+func (e *Explorer) label() string {
+	if e.entity == entityModule {
+		return "module"
+	}
+
+	return "stack"
+}
```
eliecharra: Nit, but maybe we can do a `e.entity.String()` and implement that on the custom type directly?

internal/logs/query.go
```diff
@@ -11,27 +11,69 @@ import (
...
+	Exists   bool `graphql:"exists"`
+	Finished bool `graphql:"finished"`
+	HasMore  bool `graphql:"hasMore"`
+	Messages []struct {
+		Body string `graphql:"message"`
+	} `graphql:"messages"`
+	NextToken *graphql.String `graphql:"nextToken"`
+}
+
+func queryRun[T any](ctx context.Context, e *Explorer, variables map[string]any) (*T, error) {
```
eliecharra: Is it really worth using generics here? Can we simply keep two separate functions, as the entire logic is wrapped into if statements?

eliecharra (approved):

internal/logs/query.go
tomasmik: Sure we can.

internal/logs/option.go
```diff
@@ -5,8 +5,23 @@ import "github.com/spacelift-io/spacectl/client/structs"
...
+
+const (
+	entityStack entity = iota
+	entityModule
+)
+
 type ActionOnRunState func(state structs.RunState, stackID, runID string) error
 
+// WithModule indicates that the explored run belongs to a module rather than a stack.
+func WithModule() Option {
```
mbialon: I think that passing the module as an option can be quite error-prone and not easy to figure out.

Forcing the caller to pass either `stack` or `module` option would be much cleaner.

mbialon (approved):

internal/logs/option.go
tomasmik: Didn't want to make too many changes, but fair.

Issue #381, opened 2026-02-12
[FEATURE] Ability to specify the profile when running the CLI


mohamed-haidara-cko:
## Feature Description

Ability to specify the profile when running the CLI **instead of selecting it before** running any command.

## Problem Statement

I work across multiple Spacelift organizations and I have to switch between them to perform the actions I want. This is currently a bit cumbersome as I need to select the profile before with a `spacectl profile select xxxx`.

## Proposed Solution

Similar to the AWS CLI:

- Add a new `--profile` flag (for example)  
- and env var to easily switch between profiles: `SPACELIFT_PROFILE=xxx`

## Use Cases

Any engineer working with multiple orgs can benefit from this feature.

## Additional Context

N/A

tomasmik:
It's not well documented but you can build session from env variables - https://github.com/spacelift-io/spacectl/blob/main/client/session/from_environment.go#L16

so this could be achieved using that

PR #475, opened 2026-08-08
feat: allow selecting a profile via SPACELIFT_PROFILE

derhally:
## Description

Adds `SPACELIFT_PROFILE`, which selects an existing profile by alias for a single command or shell, without changing the profile stored by `spacectl profile select`.

## Type of Change

- [x] New feature

## Changes Made

- `SPACELIFT_PROFILE` overrides the selected profile in `ProfileManager.Current()`. It is read once in `NewProfileManager` and never persisted, so `profile select` in another shell is unaffected.
- It takes precedence over credentials in the environment (`SPACELIFT_API_KEY_ID` and friends), and does not fall back to them if the profile fails: naming a profile is deliberate, whereas those variables are easy to leave set in a shell.
- An unknown alias is reported as such instead of surfacing as "no current profile is set".
- `profile select` and `profile login` warn when the variable is set, matching the existing warning from #358.
- Files: `client/session/{profile_manager,new,from_current_profile}.go`, `internal/cmd/profile/{current,select,login}_command.go`, `README.md`.

No GraphQL API features are involved, so no command versioning is needed.

## Testing

- [x] I have added/updated tests as needed
- [x] All existing tests pass

Verified against a real account that each profile resolves to its own API key, and that `~/.spacelift/config.json` is unmodified. `gofmt -s`, `go vet ./...`, `golangci-lint run` and `go test ./...` are clean.

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code where necessary
- [x] My changes generate no new warnings

## Related Issues

Related to #381, which requested exactly this. It is closed, so this PR cannot close it automatically - a maintainer may want to reopen it. The reporter noted they lacked permission to do so themselves.

---

This contribution was developed with the assistance of AI tools, and reviewed and tested before submitting.

### DIFF
--- a/client/session/new.go
+++ b/client/session/new.go
@@ -9,15 +9,30 @@ import (
 
 // New creates a session using the default chain of credentials sources:
 // first the environment, then the current credentials file.
+//
+// Naming a profile through SPACELIFT_PROFILE is an explicit choice, so it wins over credentials
+// that merely happen to be present in the environment: those are skipped entirely, rather than
+// risk authenticating as somebody other than the profile that was asked for.
 func New(ctx context.Context, client *http.Client) (Session, error) {
-	session, envErr := FromEnvironment(ctx, client)(os.LookupEnv)
-	if envErr == nil {
-		return session, nil
+	skipEnvironment := os.Getenv(EnvSpaceliftProfile) != ""
+
+	var envErr error
+	if !skipEnvironment {
+		session, err := FromEnvironment(ctx, client)(os.LookupEnv)
+		if err == nil {
+			return session, nil
+		}
+
+		envErr = err
 	}
 
 	session, fileErr := FromCurrentProfile(ctx, client)
 	if fileErr == nil {
 		return session, nil
+	}
+
+	if skipEnvironment {
+		return nil, fileErr
 	}
 
 	return nil, fmt.Errorf("could not build the session from the environment (%v) or file (%v)", envErr, fileErr)
--- a/README.md
+++ b/README.md
@@ -317,6 +317,19 @@ By default the login process is interactive, however, if that does not fit your 
 
 You can switch between account profiles by using `spacectl profile select ${MY_ALIAS}`. What this does behind the scenes is point `${HOME}/.spacelift/current` to the new location. You can also delete stored credetials for a given profile by using the `spacectl profile logout ${MY_ALIAS}` command.
 
+#### Selecting a profile per command or shell
+
+`spacectl profile select` changes the profile for every shell, since the selection is stored in the config file. When you need one terminal on a different profile - a different organization, or a different API key within the same organization - set the `SPACELIFT_PROFILE` environment variable instead:
+
+```bash
+export SPACELIFT_PROFILE=my-other-account   # for the rest of this shell
+SPACELIFT_PROFILE=my-other-account spacectl stack list   # for a single command
+```
+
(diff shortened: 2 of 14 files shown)


client/session/new.go
```diff
@@ -9,7 +9,21 @@ import (
...
 // New creates a session using the default chain of credentials sources:
 // first the environment, then the current credentials file.
+//
+// SPACELIFT_PROFILE is the exception to that order: naming a profile is an explicit choice,
+// so it wins over credentials that merely happen to be present in the environment. If the
+// named profile cannot produce a session we fail rather than falling back, to avoid silently
+// authenticating as somebody other than the profile that was asked for.
 func New(ctx context.Context, client *http.Client) (Session, error) {
+	if os.Getenv(EnvSpaceliftProfile) != "" {
+		session, err := FromCurrentProfile(ctx, client)
```
mbialon: We should not change the "current" profile selection, but introduce a new function, called something like `FromProfileOverride`, and use it here along `FromEnvironment` and `FromCurrentProfile`.

I believe it will result in a cleaner changeset, without the need for a separate validation method and reading the environment variable multiple times.

mbialon (changes requested):
Thank you for your contribution.

I believe this change could be made cleaner without the need to change the existing behaviour too much.

client/session/new.go
derhally: Thanks, yup that makes sense.  Done.

mbialon (changes requested):
I'm sorry, @derhally. I misled you with my last review.

Not changing the `(*ProfileManager).Current()` method will break some other places, like `profile export-token` command or `whoami`.

Let's get back to the original commit. My only comment about it would be that why do we need to introduce a change to `session.New()` when it's already calling `FromCurrentProfile`, which will select the correct override from the env.

Once again, sorry for the mess and confusion.

derhally:
> I'm sorry, @derhally. I misled you with my last review.
> 
> Not changing the `(*ProfileManager).Current()` method will break some other places, like `profile export-token` command or `whoami`.
> 
> Let's get back to the original commit. My only comment about it would be that why do we need to introduce a change to `session.New()` when it's already calling `FromCurrentProfile`, which will select the correct override from the env.
> 
> Once again, sorry for the mess and confusion.

No problem.  I spent more time on this.

I reverted to the original approach, as you asked. Current() handles the override again, and ValidateProfileOverride is back.

On your question about `session.New()`: the branch is there for precedence. New() tries FromEnvironment before `FromCurrentProfile`, so without it, stale SPACELIFT_API_KEY_* variables in a shell win over an explicitly set `SPACELIFT_PROFILE`. Worse, if those env credentials are merely expired, resolution falls through to the profile instead, so which account you get depends on whether the key still works. A typo'd alias also exits 0 against the env credentials rather than failing. 

The branch skips the environment entirely when the variable is set, which matches how AWS_PROFILE behaves. There are tests for both cases in client/session/new_test.go. Happy to drop it if you would still rather leave New() untouched.

While testing this I found and fixed a few related bugs:

- profile login with no alias could persist the override into currentProfileAlias, silently switching other shells.
- profile export-token printed a token for the wrong account when the alias was a typo.
- profile list and profile login gave misleading errors for an unknown alias.
- The login warning said environment variables take precedence, which is no longer true when the variable is set.

I also added a CurrentValidated() helper so the five places that resolve the profile share one guard, and added tests for each fix.

derhally:
@mbialon when you get a chance, please let me know if this needs any rework.

mbialon (approved):
LGTM

derhally:
@mbialon I missed signing my commits, fixed

